PR #1, opened 2022-03-03
Basic frontend architecture

muppi1993:
Vue app requests version information from Express server.
Version info currently hard coded into routes.js


client/src/components/VersionInfo.vue
```diff
@@ -0,0 +1,38 @@
...
+    </p>
+    <p v-for="component in versioninfo" :key="component.name">{{component.name}}
+        {{component.version}}</p>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+const API_URL = 'http://localhost:4000';
```
EmmaLRussell, line 14: I'd suggest pulling out config values like this into a an object in a separate file. Deployment is some way off, but probably good to separate out values like this which may be overwritten when deploying.

client/src/components/VersionInfo.vue
```diff
@@ -0,0 +1,38 @@
...
+      versioninfo: [],
+    };
+  },
+  mounted() {
+    this.getVersions();
+  },
+  methods: {
+    async getVersions() {
+      try {
+        const response = await fetch(`${API_URL}/version`);
```
EmmaLRussell, line 29: Is this intention to move this fetch into Vuex in a separate ticket? Generally we would avoid components making API calls directly, allowing Vuex actions to handle this, and having the components invoke the actions and display Vuex state.

EmmaLRussell (commented):
Looks good, great start!

Since you now have client and server running side-by-side I think it would probably make sense to organise the source directories to have those at the same level rather than with client nested inside the top level folder. (So could have top level `client` and `server` folders.) You might want to additionally nest those inside an `app` or `src` folder to allow for additional top level folders - you're likely to end up with at least `scripts` folder. 

In terms of testing, we'll want:
- unit tests in front end. Mostly to cover components and Vuex actions/mutations/getters. Unit testing the entry scripts etc is generally tricky so we'd tend to just roll that into browser testing. We use jest. e.g. examples in comet of [a component test with mock Vuex store](https://github.com/mrc-ide/comet/blob/main/src/app/static/tests/unit/views/about.test.ts), and a [Vuex action test](https://github.com/mrc-ide/comet/blob/main/src/app/static/tests/unit/store/actions.test.ts#L14) - this is testing an action which uses axios to do the api fetch, and we use axios-mock-adapter to mock the responses for testing. 
 
- unit testing in back end. We don't really have any examples of this yet for express apps. You might need to re-arrange your routes.js code a bit to make the functions which return responses testable (the functiona which get passed to `app.get`). I've got some slightly corresponding examples in the wodin branch [here](https://github.com/mrc-ide/wodin/pull/1/files#diff-309d22fa9b8382abb01547a87083d1026066d01a7a7cec275e03acbeb36bde3eR6), except this is complicating things a bit by using the same test code to test multiple endpoints and using the function name as parameter to that shared function. You should just be able to call the method and test that expected things happened.  
 
- browser tests - we've used Selenium in the past, and have tried out Playwright more recently. Example Playwright tests in MINT are [here](https://github.com/mrc-ide/mint/blob/master/src/app/static/src/tests/e2e/mint.etest.ts). 
 
- back end integation tests - tests of calling the running api directly. Again, we don't really have any Node examples of this. I'd expect to probably use jest and axios (or plain old fetch) to fetch from api and test that expected data is returned. 

Obviously don't need to do all these in this branch but i guess will need new tickets.

richfitz (changes requested):
considerable overlap here with Emma wrote on small quality of life things. Would be good to get a few examples of tests of each type in at the beginning so that we get the structure nailed down.

muppi1993:
I have added a new ticket to implement testing in a new PR

EmmaLRussell (commented):
Looks good to me, great to get the tests in. Might want to keep src and tests in server, as you have done on client?

README.md
```diff
@@ -1,10 +1,25 @@
...
 ```
-node index.js
+git clone git@github.com:bacpop/beebop.git
 ```
-Then go to `client/` and start the vue App:
+
+To start the Express server, go into the `app/server/` directory and run
+```
+npm install
+npm run express
```
richfitz, line 11: this fails for me:

```
$ npm run express

> beebop@1.0.0 express
> ts-node src/index.ts

sh: 1: ts-node: not found
```

Do we rely on a global ts-node installation? Or (probably better) should this be part of the package.json?

README.md
```diff
@@ -1,10 +1,25 @@
...
+```
+
+In a second terminal, go to `app/client/` and start the vue App:
 ```
+npm install
 npm run serve
-```
\ No newline at end of file
+```
```
richfitz, line 18: Indicate at this point which url you should go to (`http://localhost:8080`)

app/client/package.json
```diff
@@ -44,7 +54,18 @@
     "parserOptions": {
       "ecmaVersion": 2020
     },
-    "rules": {}
+    "rules": {},
+    "overrides": [
+      {
+        "files": [
+          "**/__tests__/*.{j,t}s?(x)",
+          "**/tests/unit/**/*.spec.{j,t}s?(x)"
```
richfitz, line 62: amazing

app/client/src/store/mutations.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  setVersions(state:any, versioninfo:object) {
```
richfitz, line 2: If this `any` is expected, could you exclude linting this line so these automatic comments below don't get distracting? If this is getting changed in an upcoming iteration don't worry

richfitz (changes requested):
This looks like a great foundation to build from. A couple of very minor things to look at noted below

app/client/src/store/mutations.ts
muppi1993: I think I will later add a type for this like here: https://github.com/mrc-ide/comet/blob/main/src/app/static/src/store/state.ts, but thought I'd do that once I have figured out how the state will actually look like.

muppi1993: Have disabled the linting in that line for now

richfitz (approved):
LGTM :shipit: 

@EmmaLRussell - do you want to merge once you're happy with it?

EmmaLRussell (approved):

PR #2, opened 2022-03-18
added docker and buildkite config files

muppi1993:


app/client/docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16
+
+WORKDIR /
+COPY ./package.json ./
+RUN npm install
+
+
+COPY . ./
```
richfitz, line 8: Better to copy into a directory above the root, then set the workdir there (drop the `WORKDIR` from above)

```
COPY  ./package.json /app/package.json
WORKDIR /app
```

then

```
COPY . /app
```

richfitz (changes requested):
A few comments here to tidy up the docker build, but this looks great. Please make the changes to both the client and server

PR #3, opened 2022-03-21
Calling python API from express server

muppi1993:
- Express server now gets version info from Flask app
- Added script to pull up beebop_py container and run both frontend components
- Express server tests now use axios mock adapter to replace flask API for testing

### DIFF
--- a/app/client/playwright.config.ts
+++ b/app/client/playwright.config.ts
@@ -1,0 +1,105 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
+import { devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+const config: PlaywrightTestConfig = {
+  testDir: './tests/e2e',
+  /* Maximum time one test can run for. */
+  timeout: 30 * 1000,
+  expect: {
+    /**
+     * Maximum time expect() should wait for the condition to be met.
+     * For example in `await expect(locator).toHaveText();`
+     */
+    timeout: 5000
+  },
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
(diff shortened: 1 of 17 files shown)


richfitz (commented):
Couple of minor comments. 

Also - it looks like we don't have any tests that run the app up against the actual docker container. @EmmaLRussell  - how does that pattern normally look (e.g., for OW or hint, assuming we actually do that)

app/server/resources/versionInfo.json
```diff
@@ -1,5 +1,5 @@
 {
-    "status":"success",
-    "errors":[],
-    "data":[{"name":"beebop","version":"0.1.0"},{"name":"poppunk","version":"2.4.0"}]
+    "data": [ { "name": "beebop", "version": "0.1.0" } ],
```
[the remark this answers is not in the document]
muppi1993: No, the tests also pass when removing the whitespace

EmmaLRussell:
> Also - it looks like we don't have any tests that run the app up against the actual docker container. @EmmaLRussell - how does that pattern normally look (e.g., for OW or hint, assuming we actually do that)

You mean testing end to end with the 'back back end' running? We normally have some sort of `run-dependencies.sh` script which would run the back back end and any other required docker containers, potentially from a docker compose file. We'd then have a suite of end-to-end tests which run when the dependencies and the app itself are running (app could be inside docker or not), make requests of the API or web app and do some basic checks. 

You probably mostly want to check that things look right in the browser, so could try Playwright tests (a bit more modern than Selenium). We don't have many examples of these, but there's a couple in in MINT (called e2e for end-to-end): 
https://github.com/mrc-ide/mint/tree/master/src/app/static/src/tests/e2e

Have a look at the MINT front end readme for a bit more on the setup for the e2e tests: 
https://github.com/mrc-ide/mint/tree/master/src/app/static#testing

app/client/playwright.config.ts
```diff
@@ -0,0 +1,105 @@
...
+    },
+
+    {
+      name: 'webkit',
+      use: {
+        ...devices['Desktop Safari'],
+      },
+    },
+
+    /* Test against mobile viewports. */
```
EmmaLRussell, line 66: I guess you can remove these commented out bits if you don't need them? I think it's ok to not target mobile browsers for an app like this.

app/client/tests/e2e/e2e.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { test, expect, Page } from '@playwright/test';
```
EmmaLRussell, line 1: Linting warning for this unused `Page` inclusion

EmmaLRussell (commented):
Playwright tests look good to me. You'll probably find as the app gets more complex that you'll want a test file for each page, so e2e.spec.ts will probably be replaced by this, but this seems like a good pattern to start with.

EmmaLRussell (approved):

richfitz (approved):

PR #4, opened 2022-04-07
Added OAuth login

muppi1993:
Added OAuth login with passport.js for Google and Github Login:
- Respective routes added to the express server
- Secrets for the config file are stored in the vault, script added to get them
- Added Login/Logout buttons to Home view
- Server integration tests use mock auth strategy to test protected routes
- Client unit tests to test home view content depending on login status


app/client/src/components/LoginPrompt.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div class=social>
+    <ul>
+      <li><a href=http://localhost:4000/login/google class="btn btn-block btn-social btn-google"><i class='fa-brands fa-google'></i> Login with Google</a></li>
```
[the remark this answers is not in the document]
richfitz: However, thinking about this - the above approach won't work if port 4000 is not exposed (which it won't be in a working app). Thoughts here @EmmaLRussell ?

richfitz (changes requested):
I think there's a little bit more required to deal with the fact that 4000 is not going to be exposed in the production app, but I have no idea what the correct way to handle this is

app/client/src/components/GreetingAndLogout.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+  <div>
+    <p>Nice to see you, {{user.data.name}}!</p>
```
EmmaLRussell, line 3: :sunglasses:

app/server/src/configApp.ts
```diff
@@ -0,0 +1,66 @@
...
+            clientSecret: config.GITHUB_CLIENT_SECRET,
+            callbackURL: '/return/github'
+        },
+            (accessToken, refreshToken, profile, cb) => {
+                return cb(null, profile);
+            }
+        )
+    );
+
+    passport.use(new MockStrategy());
```
EmmaLRussell, line 39: Is there any way to move this out of here? It's not ideal to have test code in the main class like this, but not sure if there's any way to apply passport strategies from the test code. We use Hoverfly in OrderlyWeb to mock Oauth responses from github, some approach like that might be possible?

`configApp` should certainly be unit-testable by mocking passport itself, but integration tests are trickier.

app/server/src/configApp.ts
```diff
@@ -0,0 +1,66 @@
...
+import morgan from "morgan";
+import passport from 'passport';
+import session from 'express-session';
+import PassportGoogle from 'passport-google-oauth20';
+import PassportGithub from 'passport-github';
+import MockStrategy from 'passport-mock-strategy';
+
+import config from './resources/config.json';
+
+export const configApp = (app => {
```
EmmaLRussell, line 12: Maybe rename this method to make it sound more verb-y and less noun-y?

```suggestion
export const configureApp = (app => {
```

app/server/src/routes/routes.ts
```diff
@@ -1,16 +1,94 @@
...
+    app.get('/version',
+        getVersionInfo);
+
+    app.get('/login/google',
+        passport.authenticate('google', { scope: ['profile'] }));
+
+    app.get('/login/github',
+        passport.authenticate('github', { scope: ['profile'] }));
+
+    app.get('/login/mock', passport.authenticate('mock'), (req, res) => {
```
EmmaLRussell, line 23: Again, if this is just here for testing, it would be better to move this into testing code, if possible, and it's not really testing the other routes. I think mock strategy is really a way of setting up a mocked backend to test that front end code is interacting with it as expected, not really a way of testing the server itself.

EmmaLRussell (changes requested):
Works great, buttons look good. 

Nice approach with the sercret substitution in the config file. 👍 

Could use the bootstrap social css from npm (https://www.npmjs.com/package/bootstrap-social) rather than needing to include it in the repo. We tend to pull stuff like that into main sass file e.g. https://github.com/mrc-ide/wodin/blob/main/app/static/src/scss/style.scss which you should probably set up since you're likely to have your own styles to apply soon. 

I think with vue-cli, it's just a case of including `node-sass` and `sass-loader` packages, and importing sass file into top level component e.g.  https://github.com/mrc-ide/wodin/blob/c8ed89180b284bf7d198f0451b298478929d5beb/app/static/src/app/components/basic/BasicApp.vue#L46

I don't think the integration testing approach is quite right, but not sure what would be! I don't really like having to use mock strategies in the main classes. In order to get integration and e2e tests working, I think it might be better get set up with something like [Hoverfly](https://hoverfly.io/) to mock the response from google/github so can test that the app and passport are working together as expected.

app/client/src/store/index.ts
```diff
@@ -5,6 +5,7 @@ import mutations from '@/store/mutations';
 export default createStore({
   state: {
     versions: [],
+    user: false,
```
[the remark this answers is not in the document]
EmmaLRussell: That sound very weird. Maybe try using `new Vuex.Store` with a Typescript type instead, like here: https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/fit.ts#L7

Then you can specify your state interface with `user: null | UserInfo` (will also need to specify type for UserInfo). That might help?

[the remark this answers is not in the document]
muppi1993: I have already done that when I've added the typescript interface for the state (https://github.com/bacpop/beebop/pull/4/commits/74855c80bdd8900adf8b52e961265bf805a3bd67). I had tried this with `null` as well, but it resulted in the above described situation as well.

app/client/src/store/actions.ts
```diff
@@ -2,11 +2,22 @@ import axios from 'axios';
...
 export default {
-  async getVersions({ commit } : {commit:Commit}) {
+  async getVersions({ commit }: { commit: Commit }) {
     await axios.get(`${config.server_url}/version`)
       .then((response) => {
         commit('setVersions', response.data);
       });
   },
+  async getUser({ commit }: { commit: Commit }) {
+    await axios.get(`${config.server_url}/user`)
```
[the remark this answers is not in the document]
muppi1993: created a new [ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-21) for that

.github/workflows/playwrightCI.yml
```diff
@@ -0,0 +1,44 @@
+name: playwright CI
```
EmmaLRussell, line 1: Could optionally combine jestCI and playwrightCI into testCI - would run a bit quicker but would need to dig more to see what was going wrong. I don't have a particular preference.

EmmaLRussell (approved):

richfitz (approved):

PR #6, opened 2022-07-06
Dropzone added

muppi1993:
Implemented a Drag & Drop zone to submit fasta files.
Initial processing includes the calculation of a file hash and storing hash + filename in vuex state.

To have something testable for the e2e tests, the list of processed files is currently displayed below the dropzone. This will later be replaced by the table with wasm results.

For testing, I have adopted a RootState mock like seen [here](https://github.com/mrc-ide/comet/blob/main/src/app/static/tests/mocks.ts) and adapted the existing unit tests accordingly.


app/client/src/components/DropZone.vue
```diff
@@ -0,0 +1,54 @@
...
+    <p v-for="file in results.perIsolate" :key="file.hash" class="uploaded-info">
+      {{file.filename}} {{file.hash}}
+    </p>
+    <p class="count">{{uploadedFiles}}</p>
+  </div>
+</template>
+
+<script>
+import { useDropzone } from 'vue3-dropzone';
+import { useActions, useState } from 'vuex-composition-helpers';
```
EmmaLRussell, line 19: This is a nice little library!

app/client/tests/unit/DropZone.spec.ts
```diff
@@ -0,0 +1,50 @@
+import DropZone from '@/components/DropZone.vue';
```
EmmaLRussell, line 1: Here's an approach using mocking of vue3-dropzone that would allow you to test that useDropzone had been called with the expected parameters, and that the onDrop method calls the expected action (with expected file param). Can also mock the get props methods in the return value of useDropzone so can check those are use correctly in the template:
```
// Mock the import of vue3-dropzone so we can mock its useDropzone method
// We need to do this before we import DropZone
const mockGetRootProps = jest.fn().mockReturnValue({ rprop: 'rval' });
const mockGetInputProps = jest.fn().mockReturnValue({ iprop: 'ival' });
const mockUseDropzone = jest.fn().mockReturnValue({
  getRootProps: mockGetRootProps,
  getInputProps: mockGetInputProps,
});
jest.mock('vue3-dropzone', () => ({
  useDropzone: mockUseDropzone,
}));

/* eslint-disable import/first */
import DropZone from '@/components/DropZone.vue';
import { mount } from '@vue/test-utils';
import { RootState } from '@/store/state';
import Vuex from 'vuex';
import { mockRootState } from '../mocks';

describe('Dropzone', () => {
  const processFiles = jest.fn();

  const store = new Vuex.Store<RootState>({
    state: mockRootState({
      user: {
        status: 'success',
        errors: [],
        data: {
          name: 'Jane',
          id: '543653d45',
          provider: 'google',
        },
      },
      results: {
        perIsolate: {},
      },
      uploadedFiles: 0,
    }),
    actions: {
      processFiles,
    },
  });
  const wrapper = mount(DropZone, {
    global: {
      plugins: [store],
    },
  });

  test('does a wrapper exist', () => {
    expect(wrapper.exists()).toBe(true);
  });

  it('processes dropped files', () => {
    const file = {
      name: 'sample.fa',
      text: () => Promise.resolve('ACGTGTAGTCTGACGTAAC'),
    };

    expect(mockUseDropzone).toHaveBeenCalledTimes(1);
    const { onDrop, accept } = mockUseDropzone.mock.calls[0][0];
    expect(accept).toStrictEqual(['.fa', '.fasta']);
    onDrop([file]);
    expect(processFiles).toHaveBeenCalledTimes(1);
    expect(processFiles.mock.calls[0][1]).toStrictEqual([file]);
  });

  it('sets dropzone props', () => {
    expect(wrapper.find('.dropzone').attributes('rprop')).toBe('rval');
    expect(wrapper.find('input').attributes('iprop')).toBe('ival');
  });
});
```

EmmaLRussell (commented):
I've added some suggested ways of testing more in the DropZone tests using mocking. Were you looking for a full review? Still seems to be some slightly rough edges like commented out code in the playwright config and a type called 'MyObject' so wasn't sure.. (I think MyObject might have come from my original scribbled note, but just was meant to be an example name, sorry!)

app/client/src/types.ts
```diff
@@ -1,11 +1,31 @@
+export interface MyObject {
```
hillalex, line 1: This is equivalent to `Record<string, string>` right? So no need to define an interface, can just use a type alias. Could also make it generic, and use a clearer name, I'd suggest:
```
type Dict<T> = Record<string, T>
```

app/client/src/types.ts
```diff
@@ -1,11 +1,31 @@
...
+export interface Isolate {
+    hash?: string,
+    filename?: string,
+    amr?: string,
+    cluster?: string,
+    sketch?: string
+}
+
+export interface IsolateObject {
+    [key: string]: Isolate;
```
hillalex, line 14: then this could also be omitted and `Dict<Isolate>` used intead

app/client/src/components/DropZone.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div>
+    <div v-bind='getRootProps()' class=dropzone>
+      <input v-bind='getInputProps()' />
+      <p v-if='isDragActive'>Drop the files here ...</p>
+      <p v-else>Drag 'n' drop your fasta files here, or click to select files</p>
```
hillalex, line 6: I would probably just say "Drag and drop your fasta...", but that's just my subjective preference!

richfitz (commented):
Some minor comments - I've tried to remove the ones that overlapped with Alex's review (I looked at this before the away day but never finished it!) but apologies if there are any zombie comments left

richfitz (approved):

EmmaLRussell (approved):

PR #7, opened 2022-07-12
Added webworker

muppi1993:
The webworker is for now just sending dummy results back, which get stored in Vuex state.

### DIFF
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -65,7 +65,10 @@
     "parserOptions": {
       "ecmaVersion": 2020
     },
-    "rules": {},
+    "rules": {
+      "no-shadow": "off",
+      "@typescript-eslint/no-shadow": ["error"]
+    },
     "overrides": [
       {
         "files": [
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -81,7 +81,6 @@
         },
         "rules": {
           "@typescript-eslint/no-explicit-any": "off",
-          "@typescript-eslint/no-var-requires": "off",
           "@typescript-eslint/no-empty-function": "off"
         }
       }
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -49,7 +49,8 @@
     "jest": "^27.0.5",
     "lint-staged": "^11.1.2",
     "ts-jest": "^27.0.4",
-    "typescript": "~4.5.5"
+    "typescript": "~4.5.5",
+    "vm": "^0.1.0"
   },
   "eslintConfig": {
     "root": true,
@@ -76,7 +77,9 @@
           "jest": true
         },
         "rules": {
-          "@typescript-eslint/no-explicit-any": "off"
+          "@typescript-eslint/no-explicit-any": "off",
+          "@typescript-eslint/no-var-requires": "off",
+          "@typescript-eslint/no-empty-function": "off"
         }
       }
     ]
--- a/app/client/src/store/mutations.ts
+++ b/app/client/src/store/mutations.ts
@@ -16,4 +16,11 @@ export default {
       };
     }
   },
+  setWebworkerResult(state: RootState, input: Record<string, string>) {
+    if (input.type === 'amr') {
+      state.results.perIsolate[input.hash].amr = input.result;
+    } else if (input.type === 'sketch') {
+      state.results.perIsolate[input.hash].sketch = input.result;
+    }
+  },
 };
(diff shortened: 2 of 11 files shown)


app/client/tests/unit/worker.spec.ts
```diff
@@ -0,0 +1,27 @@
+import fs from 'fs';
+
+const vm = require('vm');
```
hillalex, line 3: could this one be imported the same way as `fs`? then the `no-vars-require` rule could be left on: https://github.com/bacpop/beebop/pull/7/files#diff-b8d2936a034f3ef827722f0cd18b81d5522b816bfe559ce499e8841f2142aeafR81

app/client/src/store/mutations.ts
```diff
@@ -16,4 +16,11 @@ export default {
       };
     }
   },
+  setWebworkerResult(state: RootState, input: Record<string, string>) {
```
hillalex, line 19: missing unit test for the mutation?

hillalex (approved):
v minor comments, looks good

richfitz (commented):
This is beyond me I'm afraid, I look forward to the blog post that could get me on the page though

app/client/tests/unit/worker.spec.ts
muppi1993: I tried, but I would then get an error : `Argument of type 'Buffer' is not assignable to parameter of type 'string'` for `data` in [line 6](https://github.com/bacpop/beebop/blob/d588b3743463882384a50498956eee262798ccb1/app/client/tests/unit/worker.spec.ts#L6).
It is also working the other way round (requiring fs, importing vm), but not when both are imported...

app/client/tests/unit/actions.spec.ts
```diff
@@ -12,6 +12,23 @@ function responseSuccess(data : any) {
...
+    this.url = stringUrl;
+    this.onmessage = () => {};
+  }
+
+  postMessage(msg: any) {
+    this.onmessage({ data: msg });
+  }
+}
+
+(window as any).Worker = MockWorker;
```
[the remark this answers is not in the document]
muppi1993: All credit goes to Alex!

app/client/tests/unit/worker.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const data = fs.readFileSync('./public/worker.js');
+const script = new vm.Script(data);
+const worker: Partial<Worker> = {
+  onmessage: () => {},
+  postMessage: jest.fn(),
+};
+script.runInNewContext(worker);
```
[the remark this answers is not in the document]
muppi1993: Unfortunately webworkers are not supported by jest. When trying to generate a new one it fails with `ReferenceError: Worker is not defined`

app/client/tests/unit/worker.spec.ts
hillalex: Woah, that's odd. Let me take a look and see if I can figure out why

hillalex: Ok, this is because it's a typescript error, but if one of the packages is imported with `require` it can't perform type checking.

The fix is:

```
import fs from 'fs';
import vm from 'vm';

const data = fs.readFileSync('./public/worker.js', { encoding: 'utf8' });
```

app/client/package.json
```diff
@@ -65,7 +65,10 @@
     "parserOptions": {
       "ecmaVersion": 2020
     },
-    "rules": {},
+    "rules": {
+      "no-shadow": "off",
+      "@typescript-eslint/no-shadow": ["error"]
```
EmmaLRussell, line 70: Any particular reason to turn this rule off? It's quite a sensible one to avoid ambiguity usually.

EmmaLRussell (approved):
Looks good to me - just a query about one of the linting rules, but by all means take it out if it's holding you back!

app/client/package.json
muppi1993: I was getting an error (`already declared in the upper scope`) after introducing the enum in `types.ts`. This seems to be a [common issue](https://github.com/typescript-eslint/typescript-eslint/issues/2552#issuecomment-691694839), and the only solution I found was to disable the rule completely. This was even suggested in the [typescript-eslint documentation](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-shadow.md#how-to-use).

PR #8, opened 2022-07-20
Add wasm to webworker 

muppi1993:
Webworker should run
- sketching
- AMR prediction.

Results should be saved to vuex state.

Also adding codecov for test coverage.


muppi1993:
Code coverage is not including `worker.js`, since the way it's [imported](https://github.com/bacpop/beebop/blob/351a43b6f1273f5f37cbfa502443c7d45569f4e1/app/client/tests/unit/worker.spec.ts#L5) into tests does not allow to see coverage information on that file.

app/client/src/components/DropZone.vue
```diff
@@ -8,7 +8,7 @@
     <p> Uploaded files:</p>
     <!--only adding this temporarily to have something testable for e2e tests-->
     <div class="uploaded-info"><p v-for="file in results.perIsolate" :key="file.hash">
-      {{file.filename}} {{file.hash}} {{file.amr}} {{file.sketch}}
+      {{file.filename}} {{file.hash}} {{file.amr.slice(2, 10)}} {{file.sketch.slice(2, 4)}}
```
EmmaLRussell, line 11: These numbers are a bit magic too - rather than doing the slicing here, could you make methods to do that, with some explanatory comments

app/client/tests/unit/router.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Component is lazy loaded so need to invoke the import
+    const componentPromise = (routes[1].components.default as () => Promise<any>)();
+    componentPromise.then(
+      (component: any) => {
+        expect(component.default.name).toBe('AboutView');
+        expect(component.default.components).toHaveProperty('VersionInfo');
+        done();
+      },
+    );
+  });
```
EmmaLRussell, line 29: Could you do this like:
```suggestion
  it('About route loads About component', async () => {
        const component =  await routes[1].components.default;
        expect(component.default.name).toBe('AboutView');
        expect(component.default.components).toHaveProperty('VersionInfo');
  });
```

app/client/tests/unit/store/index.spec.ts
```diff
@@ -0,0 +1,19 @@
+import index from '@/store/index';
```
EmmaLRussell, line 1: I'm guessing these are just to keep codecov happy? I suppose rather than testing the mutation names etc it might be quicker to just expect mutations toBe the mutations imported from the mutations file?

EmmaLRussell (commented):
So the web_sketch and web_amr_prediction wasm and js files are the web assembly parts with related js - which were pre-existing and ported over from the previous app? I guess it makes sense for them to go in the `public` folder rather than `src` in that case, if they're effectively external to the source of this project, but would it be a more clean separation to make them available as a parckage through npm? (If that would work!)

Is the reason `worker.js` isn't included in codecov because it's also in `public`? Why can't it go in `src` and be compiled from ts like the other source files? Because ts wouldn't recognise the web assembly methods it calls? Oh, or is the codecov issue because you do `script.runInNewContext` to test it?

I'm struggling to run this manually - I get loads of Watchpack errors. I think this was happening before, though, so probably just an issue on my machine.

app/client/public/worker.js
```diff
@@ -1,6 +1,29 @@
...
+    var amr_result = module.make_prediction("/working/" + f.name);
+    // return result
+    postMessage({ hash: message.data.hash, type: 'amr', result: amr_result });
+  });
+
+  // same for sketch
+  WebSketch().then(module => {
+    const f = message.data.fileObject;
+    module.FS.mkdir('/working');
+    module.FS.mount(module.FS.filesystems.WORKERFS, { files: [f] }, '/working');
```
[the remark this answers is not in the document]
muppi1993: If I understand it correctly, using the WORKERFS does not actually copy the file somewhere, but instead provides the worker with read-only access to them ( see documentation here: https://emscripten.org/docs/api_reference/Filesystem-API.html#workerfs).

app/client/src/components/DropZone.vue
muppi1993: This is actually just meant for testing. When I have the final table representation of the results I won't need this anymore, just have it for now to have something testeable on the page that represents what's stored in state. I'm using the .slice() since the actual sketch and amr results are too verbose to print completely

muppi1993:
> So the web_sketch and web_amr_prediction wasm and js files are the web assembly parts with related js - which were pre-existing and ported over from the previous app? I guess it makes sense for them to go in the `public` folder rather than `src` in that case, if they're effectively external to the source of this project, but would it be a more clean separation to make them available as a parckage through npm? (If that would work!)
> 
Yes, all `web_...` files are taken from poppunk-web and amima, with the (uncompiled C++) source codes being [here](https://github.com/bacpop/pp-sketchlib/blob/master/src/web/web_sketch.cpp) and [here](https://github.com/bacpop/AMR_prediction/blob/main/amr_pred.cpp). I have no idea about npm packages, would it basically mean I just download these files into their own folder in node_modules and import from there? This would require some manual adjustments of the `.js` files, as they specify where the `.wasm` files are located, and 'default' to the location where the worker file is located.
> Is the reason `worker.js` isn't included in codecov because it's also in `public`? Why can't it go in `src` and be compiled from ts like the other source files? Because ts wouldn't recognise the web assembly methods it calls? Oh, or is the codecov issue because you do `script.runInNewContext` to test it?
> 
Yes, I think the webworker is not covered by codecov because it's not imported as other scripts but loaded with `script.runInNewContext`... It has also 0 coverage when it is placed in the `scr` folder. Also it seems that the worker is only properly working when I have the `worker.js` file in `/public` (That's also how I've seen it in different tutorials on webworkers & vue, but I couldn't find a good explanation). If it's in `/src`, I get an error: 
`loading worker from [...] was blocked because of a disallowed mime type (text/html)`
> I'm struggling to run this manually - I get loads of Watchpack errors. I think this was happening before, though, so probably just an issue on my machine.
>
Where would you get these errors, in your terminal when running `npm run serve`, or in the browsers console? I don't see any errors at the moment, but not sure I checked all relevant places?

EmmaLRussell:
> Where would you get these errors, in your terminal when running `npm run serve`, or in the browsers console? I don't see any errors at the moment, but not sure I checked all relevant places?

Yes, in the terminal, and then the app is empty in the browser. You would definitely have noticed. Don't worry about it, I'm sure it's just an issue with my machine.

EmmaLRussell (approved):

hillalex (approved):

PR #9, opened 2022-07-28
Add 'Start Analysis' button

muppi1993:
Added a button that triggers an action that submits all sketches to backend for poppunk to run.


app/client/src/store/mutations.ts
```diff
@@ -19,4 +21,10 @@ export default {
   setIsolateValue(state: RootState, input: IsolateValue) {
     state.results.perIsolate[input.hash][input.type] = input.result;
   },
+  setStatus(state: RootState, status: Record<string, string>) {
+    state.analysisStatus[status.task as keyof AnalysisStatus] = status.data;
```
richfitz, line 25: is this `as` always ok? Is there a better way to do it than `as` here?

richfitz (commented):
I don't have much I can say on this, works for me at least

.github/workflows/playwrightCI.yml
```diff
@@ -10,7 +10,6 @@ jobs:
   e2e:
 
     runs-on: ubuntu-latest
-    timeout-minutes: 20
```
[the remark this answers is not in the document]
muppi1993: Exactly, I was getting annoyed that the tmate session was aborted every 20 minutes. The actual runtime of that workflow was 5m 34s

app/client/src/store/mutations.ts
muppi1993: I am not sure, maybe @EmmaLRussell can advise on a better solution?

app/client/src/components/StartButton.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div class="start-analysis">
+    <button v-if='!analysisStatus.submitted && allSketched' class="btn btn-block btn-standard"
+    @click='runPoppunk();setStatus({ task: "submitted", data: "submitted"})'>
+      Start Analysis</button>
+    <button v-else class="btn btn-block btn-disabled">
+      Start Analysis</button>
```
EmmaLRussell, line 7: You should be able to avoid some repetition here if you just want to disable the button under that condition.
```suggestion
    <button :class="(!analysisStatus.submitted && allSketched) ? '' : 'disabled'" class="btn btn-block btn-standard"
    @click='runPoppunk();setStatus({ task: "submitted", data: "submitted"})'>
      Start Analysis</button>
```

app/client/src/components/StartButton.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div class="start-analysis">
+    <button v-if='!analysisStatus.submitted && allSketched' class="btn btn-block btn-standard"
+    @click='runPoppunk();setStatus({ task: "submitted", data: "submitted"})'>
```
EmmaLRussell, line 4: If you need a multi-line handler, it's probably more readable to put it in a method e.g. 'handleSubmit' and call that.

app/client/src/components/StartButton.vue
```diff
@@ -0,0 +1,36 @@
...
+    <!--only adding this temporarily to have something testable for e2e tests-->
+    <p class="status" v-if='analysisStatus.submitted'>{{analysisStatus}}</p>
+  </div>
+</template>
+
+<script lang='ts'>
+import { defineComponent } from 'vue';
+import { mapActions, mapMutations, mapState } from 'vuex';
+
+export default defineComponent({
```
EmmaLRussell, line 17: Is there a reason you're not using Composition API here, as you are in Dropzone?

app/client/src/store/mutations.ts
EmmaLRussell: Could make the status an enum with string values, and use those both here and when defining the keys on analysisStatus

EmmaLRussell (commented):
Just some initial comments, haven't been able to run the app yet, will try later today!

EmmaLRussell (approved):
LGTM

richfitz (approved):

PR #10, opened 2022-08-04
Add requesting job status and results from backend 

muppi1993:
Once StartButton is clicked and sketches are submitted to backend, the frontend periodically requests job statuses until all jobs finished.

When assign results are ready, they are requested and added to state.

Unfortunatley not alle tests are working yet:
- Unit tests: StartButton.spec.ts:
    - [This line ]( https://github.com/bacpop/beebop/blob/98e5a669980f1b49c1c470dae47d57581b17937d/app/client/tests/unit/components/StartButton.spec.ts#L72 )gives a ts error
```Property 'handler' does not exist on type 'string'```
however when running the test, this line is processed correctly and the test passes
    - [This test ](https://github.com/bacpop/beebop/blob/98e5a669980f1b49c1c470dae47d57581b17937d/app/client/tests/unit/components/StartButton.spec.ts#L84) is not working ( also previous line same ts error as above), even though I do not understand why it should not, since it is so similar to the previous one. Any ideas?

- End-2-end tests: Playwright tests are taking super long. For now I have to increase the timeout by 10times, will reduce later to a more sensible value once I had time to see how long they actually take.

### DIFF
--- a/app/client/src/store/actions.ts
+++ b/app/client/src/store/actions.ts
@@ -89,4 +89,10 @@ export default {
     const inter = setInterval(() => { dispatch('getStatus'); }, 1000);
     commit('setStatusInterval', inter);
   },
+  async submitData(context: ActionContext<RootState, RootState>) {
+    const { dispatch, commit } = context;
+    dispatch('runPoppunk');
+    commit('setSubmitStatus', 'submitted');
+    dispatch('startStatusPolling');
+  },
 };
--- a/app/client/src/store/actions.ts
+++ b/app/client/src/store/actions.ts
@@ -61,4 +61,40 @@ export default {
       commit('setStatus', { task: 'network', data: 'submitted' });
     });
   },
+  async getStatus({ commit, state }: { commit: Commit, state: RootState }) {
+    await axios.post(
+      `${config.server_url}/status`,
+      { hash: state.projectHash },
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    ).then((response) => {
+      const prev = { ...state.analysisStatus };
+      if (response.data.data.microreact !== prev.microreact) {
+        commit('setStatus', { task: 'microreact', data: response.data.data.microreact });
+      }
+      if (response.data.data.network !== prev.network) {
+        commit('setStatus', { task: 'network', data: response.data.data.network });
+      }
+      if (response.data.data.assign !== prev.assign) {
+        commit('setStatus', { task: 'assign', data: response.data.data.assign });
+      }
+    });
+  },
+  async getAssignResult({ commit, state }: { commit: Commit, state: RootState }) {
+    await axios.post(
+      `${config.server_url}/assignResult`,
+      { projectHash: state.projectHash },
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    )
+      .then((response) => {
+        commit('setClusters', response.data);
+      });
+  },
 };
--- a/app/client/src/store/actions.ts
+++ b/app/client/src/store/actions.ts
@@ -1,10 +1,23 @@
 import axios from 'axios';
-import { Commit } from 'vuex';
+import { Commit, ActionContext } from 'vuex';
 import config from '@/resources/config.json';
 import { Md5 } from 'ts-md5/dist/md5';
 import { RootState } from '@/store/state';
 
 axios.defaults.withCredentials = true;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+async function axiosJsonPost(endpoint: string, body: Record<string, any>) {
+  return axios.post(
+    `${config.server_url}/${endpoint}`,
+    body,
+    {
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    },
+  );
+}
 
 export default {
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: app/client/src/store/actions.ts [hunk 1]
```diff
@@ -61,4 +61,40 @@ export default {
...
     });
   },
+  async getStatus({ commit, state }: { commit: Commit, state: RootState }) {
+    await axios.post(
+      `${config.server_url}/status`,
+      { hash: state.projectHash },
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
```
EmmaLRussell, line 71: To avoid repetition, you might want to pull out doing a post with json header, with base url,  into a helper method, or consider using an `apiService`-type class as we do in other projects e.g. https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/apiService.ts

Could be in another ticket if you use a full apiService class.

app/client/src/store/actions.ts
```diff
@@ -61,4 +61,40 @@ export default {
...
+      },
+    ).then((response) => {
```
EmmaLRussell, line 73: Any error handling? Should it display an error if the request fails?

app/client/src/store/actions.ts
```diff
@@ -61,4 +61,40 @@ export default {
...
+      const prev = { ...state.analysisStatus };
+      if (response.data.data.microreact !== prev.microreact) {
+        commit('setStatus', { task: 'microreact', data: response.data.data.microreact });
+      }
+      if (response.data.data.network !== prev.network) {
+        commit('setStatus', { task: 'network', data: response.data.data.network });
+      }
+      if (response.data.data.assign !== prev.assign) {
+        commit('setStatus', { task: 'assign', data: response.data.data.assign });
+      }
```
EmmaLRussell, line 83: Would there be any downside to just committing the whole thing every time? It would simplify the code if `setStatus` could just take the whole `response.data.data` object and assign it to `state.analysisStatus`

app/client/src/store/actions.ts
```diff
@@ -61,4 +61,40 @@ export default {
...
+    await axios.post(
+      `${config.server_url}/status`,
+      { hash: state.projectHash },
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    ).then((response) => {
+      const prev = { ...state.analysisStatus };
```
EmmaLRussell, line 74: I don't think you really need to do this expansion here, doesn't look like you're editing `prev` at all.

app/client/src/store/actions.ts
muppi1993: Have added a helper function for now, but will work on the api service class in the next ticket.

app/client/src/store/actions.ts
muppi1993: Will leave that to be handled in the api service class

app/client/src/store/actions.ts
muppi1993: I just implemented it like this because I was using the watchers. They would not recognise which values were changed when I overwrite all statuses every time and fire even if a status is the same as before. have changed that now since I removed the watchers.

app/client/src/store/actions.ts
```diff
@@ -47,54 +60,33 @@ export default {
...
-    });
+  async getStatus(context: ActionContext<RootState, RootState>) {
+    const { state, dispatch, commit } = context;
+    await axiosJsonPost('status', { hash: state.projectHash })
+      .then((response) => {
+        if (response.data.data.assign === 'finished' && state.analysisStatus.assign !== 'finished') {
+          dispatch('getAssignResult');
+        }
+        commit('setAnalysisStatus', response.data.data);
+        if ((response.data.data.network === 'finished' || response.data.data.network === 'failed') && (response.data.data.microreact === 'finished' || response.data.data.microreact === 'failed')) {
```
EmmaLRussell, line 76: I'm seeing status requests continue to fire even when microreact and network are showing as 'finished'. I think that's probably because I tested if I could still click the 'Start Analysis' button after it had started running, so probably the first interval is still firing and only the second one got cleared.

Update: yes, I tried again and just clicked the button once, and it stopped polling when expected.

EmmaLRussell (commented):
Had some trouble getting this running locally - finally worked out that the easiest way to do that was to run `./scripts/run_test` to get all the deps running. Might be worth noting that in the readme?

Looks good!

The Start Analysis button still seems to be enabled while the analysis is running. But it shouldn't be, I don't think?

I think the idea is that the results displayed will be updated more fully as the analysis runs, but I wonder if there should be some visual animated feature (e.g. spinner) to indicate that things are happening, and make it obvious when they stop. Maybe there's other tickets for that already?

muppi1993:
> I think the idea is that the results displayed will be updated more fully as the analysis runs, but I wonder if there should be some visual animated feature (e.g. spinner) to indicate that things are happening, and make it obvious when they stop. Maybe there's other tickets for that already?

I already have a status bar component ready. Will add this in a later ticket.

richfitz (approved):
There's a bunch new in here for me so I can't really meaningfully comment on the code despite having read through it.

However, it seems to run really well :)

EmmaLRussell (approved):
Thanks, looks great

PR #11, opened 2022-08-25
Add apiService class to client

muppi1993:
Added an API service class adapted from [here]( https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/apiService.ts) and changed the mutation's api calls.

Had to make some changes to the shape of the state and to the tests accordingly.

### DIFF
--- a/app/client/src/apiService.ts
+++ b/app/client/src/apiService.ts
@@ -1,5 +1,4 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable no-console */
 import axios, { AxiosError, AxiosResponse } from 'axios';
@@ -26,13 +25,12 @@ export interface API<S, E> {
     withError: (type: E) => API<S, E>
     withSuccess: (type: S) => API<S, E>
     ignoreErrors: () => API<S, E>
+    ignoreSuccess: () => API<S, E>
 
     get<T>(url: string): Promise<void | ResponseWithType<T>>
 }
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 type OnError = (failure: ResponseFailure) => void;
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 type OnSuccess = (success: ResponseSuccess) => void;
 
 export class APIService<S extends string, E extends string> implements API<S, E> {
@@ -43,6 +41,8 @@ export class APIService<S extends string, E extends string> implements API<S, E>
     }
 
     private _ignoreErrors = false;
+
+    private _ignoreSuccess = false;
 
     static getFirstErrorFromFailure = (failure: ResponseFailure) => {
       if (failure.errors.length === 0) {
@@ -72,6 +72,11 @@ export class APIService<S extends string, E extends string> implements API<S, E>
 
     ignoreErrors = () => {
       this._ignoreErrors = true;
+      return this;
+    };
+
+    ignoreSuccess = () => {
+      this._ignoreSuccess = true;
       return this;
     };
 
@@ -118,7 +123,7 @@ export class APIService<S extends string, E extends string> implements API<S, E>
       if (this._onError == null && !this._ignoreErrors) {
         console.warn(`No error handler registered for request ${url}.`);
       }
-      if (this._onSuccess == null) {
+      if (this._onSuccess == null && !this._ignoreSuccess) {
         console.warn(`No success handler registered for request ${url}.`);
       }
     }
--- a/app/client/src/apiService.ts
+++ b/app/client/src/apiService.ts
@@ -11,7 +11,7 @@ export interface ResponseWithType<T> extends ResponseSuccess {
 }
 
 export function isAPIError(object: any): object is BeebopError {
-  return typeof object.error === 'string'
+  return object && typeof object.error === 'string'
         && (object.detail === undefined || typeof object.detail === 'string');
 }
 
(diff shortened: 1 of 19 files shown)


app/client/src/apiService.ts
```diff
@@ -0,0 +1,139 @@
...
+import axios, { AxiosError, AxiosResponse } from 'axios';
+import { Commit, ActionContext } from 'vuex';
+import { ResponseSuccess, ResponseFailure, BeebopError } from './types';
+
+export interface ResponseWithType<T> extends ResponseSuccess {
+    data: T
+}
+
+export function isAPIError(object: any): object is BeebopError {
+  return typeof object.error === 'string'
```
richfitz, line 14: do you need `object &&` here as for `isAPIResponseFailure`?

app/client/src/store/actions.ts
```diff
@@ -3,34 +3,25 @@ import { Commit, ActionContext } from 'vuex';
...
-  );
-}
-
 export default {
-  async getVersions({ commit }: { commit: Commit }) {
-    await axios.get(`${config.server_url}/version`)
-      .then((response) => {
-        commit('setVersions', response.data);
-      });
+  async getVersions(context: ActionContext<RootState, RootState>) {
```
richfitz, line 14: this is nice

scripts/run_test
```diff
@@ -4,7 +4,7 @@ NETWORK=beebop_nw
 VOLUME=beebop-storage
 NAME_REDIS=beebop-redis
 NAME_API=beebop-py-api
-API_BRANCH=bacpop-39
+API_BRANCH=bacpop-41
```
richfitz, line 7:
```suggestion
API_BRANCH=main
```

https://github.com/bacpop/beebop_py/pull/11 is merged now

richfitz (approved):
LGTM. Emma and I discussed that we might pull this client into its own package which would would reduce boilerplate in the future, but that's a job for later

app/client/src/apiService.ts
muppi1993: Yes, I think that makes sense so it does not return an error in case the object is `null`

app/client/src/apiService.ts
```diff
@@ -0,0 +1,139 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
EmmaLRussell, line 2: I don't think you need this exclusion

app/client/src/apiService.ts
```diff
@@ -0,0 +1,139 @@
...
+export interface API<S, E> {
+
+    withError: (type: E) => API<S, E>
+    withSuccess: (type: S) => API<S, E>
+    ignoreErrors: () => API<S, E>
+
+    get<T>(url: string): Promise<void | ResponseWithType<T>>
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
EmmaLRussell, line 33: I don't think this is needed either..

app/client/src/apiService.ts
```diff
@@ -0,0 +1,139 @@
...
+type OnError = (failure: ResponseFailure) => void;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
EmmaLRussell, line 35: ..or this.. 
(This is probably tidy-up we can do in the wodin class too!)

app/client/src/store/actions.ts
```diff
@@ -52,37 +43,48 @@ export default {
...
     commit('setProjectHash', phash);
     Object.keys(state.results.perIsolate).forEach((element) => {
       jsonSketches[element] = JSON.parse(state.results.perIsolate[element].sketch as string);
     });
-    await axiosJsonPost('poppunk', { projectHash: phash, sketches: jsonSketches })
-      .then(() => {
-        commit('setAnalysisStatus', { assign: 'submitted', microreact: 'submitted', network: 'submitted' });
-      });
+    const response = await api(context)
+      .withError('addError')
```
EmmaLRussell, line 56: I think this will currently show a warning in the console since you're not providing a mutation to run on success. If this is intentional, and there's no mutation to do with the data on some success responses, you might want to include an `ignoreSuccess` method on apiSerivce for these cases, as we have in [hint](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/apiService.ts#L20) (and check for that in `verifyHandlers`)

app/client/src/types.ts
```diff
@@ -45,6 +41,21 @@ export type AnalysisStatus = {
...
+export interface ResponseFailure {
+    status: 'failure';
+    data: null;
+    errors: BeebopError[];
+}
+
+export interface ResponseSuccess<T=unknown> {
+    status: 'success';
+    data: T;
+    errors: null;
```
EmmaLRussell, line 60: What was the reason for changing this from the original?:
```
export interface ResponseSuccess {
    status: "success";
    data: unknown;
    errors: null;
}
```

app/client/src/views/HomeView.vue
```diff
@@ -1,12 +1,12 @@
 <template>
   <div class='home'>
     <h1>Welcome to beebop!</h1>
-    <div v-if='user' class='auth'>
-      <LoginPrompt v-if='!loggedIn' />
-      <GreetingAndLogout v-if='loggedIn' />
+    <div class='auth'>
```
EmmaLRussell, line 4: I'm a bit unclear about how the changes in this view relate to the apiService?

EmmaLRussell (commented):
Great - good to get this in before the app gets any more complicated!

app/client/src/views/HomeView.vue
muppi1993: Previously I had commited the response object (with success, data, errors) in total to the user state. I first checked with the computed property `loggedIn`, whether the status was 'success' and the user was actually logged in. Since that check is now done by the api class, I do only have a user saved in state when this was a successful request. So `loggedIn` became redundant, since the user state means the same now.

app/client/src/types.ts
muppi1993: I once wanted to use something like ResponseSuccess<ClusterInfo>, but have discarded this. Will change this back to previous version

EmmaLRussell (approved):

PR #12, opened 2022-08-31
Add Status bar 

muppi1993:
Added a status bar to visualise the analysis progress.
The progress is given in percentage of jobs that have finished (regardless of their duration). 
In the future this could be a bit more sophisticated, but for MVP this should suffice?


app/client/src/components/ProgressBar.vue
```diff
@@ -0,0 +1,30 @@
...
+import { BProgress, BProgressBar } from 'bootstrap-vue-3';
+
+export default {
+  name: 'ProgressBar',
+  computed: {
+    ...mapGetters([
+      'analysisProgress',
+    ]),
+    animated() {
+      return this.analysisProgress as unknown as number !== 1;
```
EmmaLRussell, line 22: This double cast looks a little bit suspect!

app/client/src/store/getters.ts
```diff
@@ -0,0 +1,15 @@
+import { AnalysisType } from '@/types';
+import { RootState } from '@/store/state';
+
+export default {
+  analysisProgress(state: RootState) {
```
EmmaLRussell, line 5: We'd normally use the `GetterTree` vuex interface for getters.  [Example here](https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/store/modelFit/getters.ts). Not sure if that would help with your type issues when mapping from the component though.

EmmaLRussell (commented):
The status bar looks good!

Just a suggestion, but since there seems to be quite a long delay between hitting 'Start Analysis' and anything apparenlty happening on the status bar, I wonder if it would be a little clearer to have a message like "Completed x of y jobs" rather than (or as well as) the sporadically upating percentage value, which is a little misleading. I think the user would expect % to update fairly smoothly, but would understand "Completed x of y.. " message would get discrete updates. Could still show status in the progress bar as a proportion of work done though. Maybe the 'Completed..' message could go under the progress bar, so that it would be easy to initialise with 'Completed 0 of n jobs' immediately (so the user can see that something is happening), and could still show completed % on the bar itself?

app/client/src/components/ProgressBar.vue
muppi1993: When removing the `as unknown` I get the following TS Error:
```
Conversion of type 'Computed' to type 'number' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```
Seems that it cannot infer the return type from the mapped getter, even though I have specified it in getters.ts.
This happens also after using the GetterTree interface. Is it ok to just leave it like this, or should I rather try to import the getter in a different way that provides type information?

EmmaLRussell: OK, I think you might just need to declare your component using `defineComponent`, so 
```
import { defineComponent } from 'vue';

export default defineComponent({
  name: 'ProgressBar',
..etc...
 components: {
    BProgress,
    BProgressBar,
  },
});
```

app/client/src/components/ProgressBar.vue
```diff
@@ -0,0 +1,35 @@
...
+import { defineComponent } from 'vue';
+
+export default defineComponent({
+  name: 'ProgressBar',
+  computed: {
+    ...mapGetters([
+      'analysisProgress',
+    ]),
+    animated(): boolean {
+      return (this.analysisProgress as Record<string, number>).progress !== 1;
```
EmmaLRussell, line 27: This seemed to work for me without the case
```suggestion
      return this.analysisProgress.progress !== 1;
```

EmmaLRussell (commented):
I like the new job label, but not sure it's always showing the right number. Should there not be one job per file?
Screenshot from 2022-09-02 16-25-55

PR #13, opened 2022-09-08
Add results table

muppi1993:
Added the results table that displays:
- the webworker results (sketching successfull, AMR results with tooltip) once they are processed
- after clicking the submit button the job statuses for cluster assignment and microreact & network visualisations appear.
- once results are finished, clusters appear in the table. Microreact/Network only display a checkmark so far, this will be replaced in the next PR by buttons to download the results.
Once clusters are available, the table sorts by cluster and microreact/network cells from the same cluster merge (since we will only have 1 button per cluster for these).

I am a bit unsure about the watchers. Without them, the table content would not update, but I am wondering if this approach is a bit computaionally expensive? Do you have any ideas to make that more efficient?

The utils functions might seem a bit complicated. Since the raw AMR results are not very intuitive to interpret, John and I had discussed approaches to translate them into a verbal representation and also to convert them into colours (light for low probabilities, darker for high probabilities). Since the translation differs by antibiotic (due to the different models/ model performances), the functions are so long and heavy on if/else statements... Maybe this could be moved to the C++/wasm code to have a clearer separation between the research side and the website code?

### DIFF
--- a/app/client/src/components/ResultsTable.vue
+++ b/app/client/src/components/ResultsTable.vue
@@ -51,28 +51,25 @@
 import { defineComponent } from 'vue';
 import { mapState } from 'vuex';
 import { VBTooltip } from 'bootstrap-vue-3';
-import { addRowspan, getRGB, verbalProb } from '../utils';
+import {
+  addRowspan, getRGB, tooltipLine,
+} from '../utils';
 
 export default defineComponent({
   name: 'ResultsTable',
   directives: {
     'b-tooltip': VBTooltip,
   },
-  data() {
-    return {
-      tableData: this.generateTableData(),
-    };
-  },
   methods: {
     getRGB,
-    verbalProb,
+    tooltipLine,
     getTooltipText(data: Record<string, number>) {
       const firstLine = 'Probability of resistance to:';
-      const pen = ['Penicillin: ', this.verbalProb(data.Penicillin, 'Penicillin'), ' <small>(', data.Penicillin, ')</small>'].join('');
-      const chlor = ['Chloramphenicol: ', this.verbalProb(data.Chloramphenicol, 'Chloramphenicol'), ' <small>(', data.Chloramphenicol, ')</small>'].join('');
-      const ery = ['Erythromycin: ', this.verbalProb(data.Erythromycin, 'Erythromycin'), ' <small>(', data.Erythromycin, ')</small>'].join('');
-      const tetra = ['Tetracycline: ', this.verbalProb(data.Tetracycline, 'Tetracycline'), ' <small>(', data.Tetracycline, ')</small>'].join('');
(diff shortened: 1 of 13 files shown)


app/client/src/components/ResultsTable.vue
```diff
@@ -0,0 +1,134 @@
...
+      if (this.results) {
+        const samples = this.results.perIsolate;
+        const items: Array<Record<string, string>> = [];
+        Object.keys(samples).forEach((sample) => {
+          items.push({
+            Hash: samples[sample].hash,
+            Filename: samples[sample].filename,
+            Sketch: (samples[sample].sketch) ? '\u2714' : 'processing',
+            AMR: samples[sample].amr ? samples[sample].amr : 'processing',
+            Cluster: this.submitStatus === 'submitted' ? this.getCluster(sample) : '',
```
hillalex, line 98: One issue here is that if the user submits an analysis but then adds more fasta files, the new files show up with ticks even though they haven't been analysed yet. I don't know what the desired UI is here: I guess the easiest thing would be to reset the analysis when new files are added, but probably the better thing would be to have file level statuses that store whether a particular file has been analysed, so that new files can be added and then submitted for analysis.

To replicate: 1. add fasta files 2. press "start analysis" 3. add more fasta files

Having individual file statuses would also allow the progress bar to be more fine-grained, could increment it every time a single file's status was updated rather than it having to wait for all files to be processed to progress.

hillalex (changes requested):
This is looking really slick :)

The performance of the `tableData` generating code seems fine to me, so I wouldn't worry about the being computationally expensive. It should be a computed property rather than a watcher though!

Some comments about the behaviour - need to handle the case in which more files are added after some have already been analysed.

A point unrelated to the changes here: the colour of the "start analysis" button should really be green when enabled; grey is usually reserved for buttons or links that are disabled.

app/client/src/components/ResultsTable.vue
```diff
@@ -0,0 +1,134 @@
...
+      return null;
+    },
+  },
+  computed: {
+    ...mapState(['results', 'submitStatus', 'analysisStatus']),
+  },
+  mounted() {
+    this.tableData = this.generateTableData();
+  },
+  watch: {
```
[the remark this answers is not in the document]
muppi1993: Amazing, this is so much easier than using the watchers! Also simplifies the tests :)

app/client/src/components/ResultsTable.vue
muppi1993: Since the analyses in the backend (cluster, microreact & network) do run with all files at once in a batch, adding new files later would mean we would have to rerun everything with all files (can't just combine old result with new one). So it is not intended to add more files after clicking the submit button. In the mockups the view was shifted after hitting submit, so that the dropzone would disappear and no further file submission would be possible. I have added that to the conditional rendering of the dropzone component that it is not available after submission.

hillalex (approved):

PR #14, opened 2022-09-12
Towards dockerised deployment

richfitz:
Basic working dockerised app (minus auth) with instructions for running in the README.

Approach:
- 2 containers, the backend api and the nginx proxy which also serves the static files
- ~~config for the front-end trivial because no secrets, just use a webpack alias to point to development or production config~~
I spoke too soon here because of course prod and local docker need to be different, but since the static files are built into the docker image this approach would mean that we would need separate images for prod and local. For now I've just got it building using the prodution config, but we should rethink this a bit/probably follow the patterm of just copying the correct config into the container after deployment.
- config for the server uses secrets so can't be baked into the docker image; at the moment it has to be created on disk (using the `decrypt_config` script which populates with secrets from vault in a rather brittle way) and then copied into the running container. I've set up a real GitHub OAuth app, but not a Google one yet.

This has been deployed at  https://beebop.dide.ic.ac.uk (with a self-signed certificate) using

1.  `./scripts/decrypt_config production`
1. `./scripts/run_docker beebop.dide.ic.ac.uk`


docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    depends_on:
+      - beebop-redis
+    command:
+      rqworker
+    volumes:
+      - beebop-storage:/beebop/storage
+    environment:
+      - REDIS_HOST=beebop-redis
+volumes:
+  beebop-storage:
```
muppi1993, line 39: Is this volume empty? When trying out the deployed app the backend analyses won't run. What I suspect is that the database is maybe missing in the volume (We previously [downloaded the db](https://github.com/bacpop/beebop/blob/e485e7625e45d187ad9d00d308b0063ee49e67fa/scripts/run_test#L12) into the volume with [this script](https://github.com/bacpop/beebop_py/blob/da6bc1f11181c47c7a64343b09cd35c1f8695410/scripts/download_db))? 
Is there a place where the logs of beebop-py-api can be viewed?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  beebop-py-api:
+    image: mrcide/beebop-py:main
+    depends_on:
+      - beebop-redis
+    volumes:
+      - beebop-storage:/beebop/storage
+    environment:
+      - REDIS_HOST=beebop-redis
+      - STORAGE_LOCATION=./storage
+      - DB_LOCATION=./storage/GPS_v4_references
```
muppi1993, line 25: Should we use the full db instead of reference db for deployment?

EmmaLRussell (commented):
When I run this up locally and attempt to login, it redirects to beebop.dide.ic.ac.uk - I assume this is expected?

README.md
```diff
@@ -32,6 +32,31 @@ To start all required components, run:
 
 The website can be viewed at http://localhost:8080/ . You can stop the application with `./scripts/stop_test`.
 
+## Deploying with docker
+
+Make sure that all docker images exist (`./proxy/docker/build`, `./app/server/docker/build`).
```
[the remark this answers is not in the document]
richfitz: Is this even needed if buildkite is building the images for us (that looks setup)

app/server/docker/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eu
+
+PATH_CONFIG=/app/src/resources/config.json
+echo "Waiting for config at $PATH_CONFIG"
+while [ ! -e $PATH_CONFIG ]; do
+  sleep 1
+done
+
+sleep 1
+npm run express
```
richfitz, line 11: Is this a safe way of running things in prod, or should we add another ticket for running only the built version?

app/server/src/index.ts
```diff
@@ -12,11 +12,11 @@ configureApp(app)
 
 router(app)
 
-if (process.env.BEEBOP_TEST == "true"){
+if (process.env.BEEBOP_TEST == "true") {
```
richfitz, line 15: `===` too?
```suggestion
if (process.env.BEEBOP_TEST === "true") {
```

scripts/decrypt_config
```diff
@@ -1,13 +1,22 @@
...
+FILE_CLEAR_SERVER=app/server/src/resources/config.json
 
 export GITHUB_ID=$(vault read -field=clientid secret/beebop/auth/github)
 export GITHUB_SECRET=$(vault read -field=secret secret/beebop/auth/github)
+export PROD_GITHUB_ID=$(vault read -field=clientid secret/beebop/auth/production/github)
+export PROD_GITHUB_SECRET=$(vault read -field=secret secret/beebop/auth/production/github)
+export PROD_GOOGLE_ID=1234
+export PROD_GOOGLE_SECRET=1234
 export GOOGLE_ID=$(vault read -field=clientid secret/beebop/auth/google)
 export GOOGLE_SECRET=$(vault read -field=secret secret/beebop/auth/google)
```
richfitz, line 19: this part should also read out the certs from the vault - see /secret/beebop/ssl in the centre vault

proxy/ssl/key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
richfitz, line 1: can you add the generation script here so we know that these are just self-signed?

README.md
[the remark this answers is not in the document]
hillalex: This isn't needed unless you want to build an image locally from local changes

app/server/docker/entrypoint.sh
hillalex: It's not. We should build and use pm2 to serve. Can do that on this branch

docker-compose.yml
hillalex: I figured a bunch of this would be wrong, I just blindly copied this from the existing local run script! Perhaps we can have a call so you can explain how the db/redis works :)

docker-compose.yml
hillalex: See comment above, let's have a call as I have no idea how any of this works!

hillalex:
> When I run this up locally and attempt to login, it redirects to beebop.dide.ic.ac.uk - I assume this is expected?

Hmm, no. The proxy entrypoint script *should* use localhost by default. When running with just `./run_docker` you should see a message in the proxy logs like "We will listen on ports 80 (http) and 443 (https) with hostname localhost". Or, if you run it up with `./run-docker beebop.dide.ic.ac.uk` you see "We will listen on ports 80 (http) and 443 (https) with hostname beebop.dide.ic.ac.uk"

proxy/docker/push
```diff
@@ -10,7 +10,7 @@ HERE=$(dirname $0)
 docker tag $TAG_SHA $TAG_BRANCH
 docker push $TAG_BRANCH
 
-if [ $GIT_BRANCH == "main" ]; then
+if [ $GIT_BRANCH == "master" ]; then
```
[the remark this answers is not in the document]
hillalex: Yeah not sure why that changed, probably a copy paste mistake

hillalex:
> > When I run this up locally and attempt to login, it redirects to beebop.dide.ic.ac.uk - I assume this is expected?
> 
> Hmm, no. The proxy entrypoint script _should_ use localhost by default. When running with just `./run_docker` you should see a message in the proxy logs like "We will listen on ports 80 (http) and 443 (https) with hostname localhost". Or, if you run it up with `./run-docker beebop.dide.ic.ac.uk` you see "We will listen on ports 80 (http) and 443 (https) with hostname beebop.dide.ic.ac.uk".

Oh, sorry, completely misread. Yes login will redirect to prod, since the oAuth app has it's redirect url set to the prod url!

muppi1993 (commented):
Looks good to me, at least the bit I understand. I'm looking forward to when everything comes together with the components that are still missing in this version :)

EmmaLRussell (approved):

PR #15, opened 2022-09-14
Add Buttons to get visualisation results 

muppi1993:
The results table now has buttons to
- download microreact and network files as zipped folders
- generate a microreact URL
Once the microreact URL has been generated, the button turns into a link that redirects to the microreact project.

(At the moment the microreact output/ URL is not working properly - despite files/  a valid URL being generated, microreact.org will show 'An unexpected error has occurred.' for most clusters. This also happens when downloading the files and uploading to microreact. I suspect this is an issue with PopPUNK rather than beebop - needs further investigation)

### DIFF
--- a/app/client/src/components/DownloadZip.vue
+++ b/app/client/src/components/DownloadZip.vue
@@ -2,7 +2,7 @@
   <button
   @click='onClick()'
   class="btn btn-block btn-standard btn-download">
-    Download files as .zip
+    Download zip file
   </button>
 </template>
 
@@ -11,7 +11,7 @@ import { defineComponent } from 'vue';
 import { mapGetters, mapState, mapActions } from 'vuex';
 
 export default defineComponent({
-  name: 'DownloadMicroreactZip',
+  name: 'DownloadZip',
   props: ['type', 'cluster'],
   methods: {
     ...mapActions(['getZip']),
--- a/app/client/src/components/DownloadZip.vue
+++ b/app/client/src/components/DownloadZip.vue
@@ -1,0 +1,29 @@
+<template>
+  <button
+  @click='onClick()'
+  class="btn btn-block btn-standard btn-download">
+    Download files as .zip
+  </button>
+</template>
+
+<script lang='ts'>
+import { defineComponent } from 'vue';
+import { mapGetters, mapState, mapActions } from 'vuex';
+
+export default defineComponent({
+  name: 'DownloadMicroreactZip',
+  props: ['type', 'cluster'],
+  methods: {
+    ...mapActions(['getZip']),
+    async onClick() {
+      this.getZip({ type: this.type, cluster: this.cluster });
+    },
+  },
+  computed: {
+    ...mapGetters([
+      'uniqueClusters',
+    ]),
+    ...mapState(['projectHash']),
+  },
+});
+</script>
--- a/app/client/src/components/GenerateMicroreactURL.vue
+++ b/app/client/src/components/GenerateMicroreactURL.vue
@@ -10,15 +10,15 @@
       </button>
 
       <button
-        v-if='microreactToken && !results.perCluster[cluster]?.microreactURL'
-        @click='getMicroreactURL(cluster)'
+        v-if='tokenAvailable'
+        @click='buildMicroreactURL(cluster)'
         class='btn btn-block btn-standard btn-download'
       >
         Generate Microreact URL
       </button>
 
       <a
-        v-if='microreactToken && results.perCluster[cluster]?.microreactURL'
+        v-if= 'URLgenerated'
         :href='results.perCluster[cluster]?.microreactURL'
         class='btn btn-block btn-standard btn-download'
         target='_blank'
@@ -80,9 +80,9 @@
 
 <script lang='ts'>
 import { defineComponent } from 'vue';
-import { mapState, mapActions, mapMutations } from 'vuex';
+import { mapState, mapActions } from 'vuex';
 import Modal from '@/components/Modal.vue';
-import { BeebopError } from '../types';
+import { BeebopError, Errors } from '../types';
 
 export default defineComponent({
   name: 'GenerateMicroreactURL',
@@ -97,8 +97,7 @@ export default defineComponent({
     };
   },
   methods: {
-    ...mapActions(['getMicroreactURL']),
-    ...mapMutations(['setToken']),
+    ...mapActions(['buildMicroreactURL']),
     showModal() {
       this.isModalVisible = true;
     },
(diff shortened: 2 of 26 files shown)


app/client/src/components/DownloadZip.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <button
+  @click='onClick()'
+  class="btn btn-block btn-standard btn-download">
+    Download files as .zip
```
LekanAnni, line 5: Could this be `Download zip files` ?

app/client/src/components/DownloadZip.vue
```diff
@@ -0,0 +1,29 @@
...
+  </button>
+</template>
+
+<script lang='ts'>
+import { defineComponent } from 'vue';
+import { mapGetters, mapState, mapActions } from 'vuex';
+
+export default defineComponent({
+  name: 'DownloadMicroreactZip',
```
LekanAnni, line 14:
```suggestion
  name: 'DownloadZip',
```

LekanAnni: Name should correspond with file name.

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+    },
+  },
+  computed: {
+    ...mapState(['results', 'microreactToken', 'errors']),
+    tokenWasWrong() {
+      if (this.errors.some((e: BeebopError) => e.error === 'Wrong Token')) {
+        return true;
+      }
+      return false;
+    },
```
LekanAnni, line 120:
```suggestion
    tokenWasWrong() {
     return this.errors.some((e: BeebopError) => e.error === 'Wrong Token');
    },
```

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+    ...mapMutations(['setToken']),
+    showModal() {
+      this.isModalVisible = true;
+    },
+    closeModal() {
+      this.isModalVisible = false;
+    },
+    saveToken() {
+      this.setToken(this.token);
+      this.getMicroreactURL(this.cluster);
```
LekanAnni, line 110:
```suggestion
      this.buildMicroReactURL(this.cluster, this.token);
```

You really don't need to set token in here, you could do it in buildMicroReactURL, method should be renamed as it is not returning anything.

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+  },
+  data() {
+    return {
+      isModalVisible: false,
+      token: null,
+    };
+  },
+  methods: {
+    ...mapActions(['getMicroreactURL']),
+    ...mapMutations(['setToken']),
```
LekanAnni, line 101:
```suggestion

```

delete

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+    Modal,
+  },
+  data() {
+    return {
+      isModalVisible: false,
+      token: null,
+    };
+  },
+  methods: {
+    ...mapActions(['getMicroreactURL']),
```
LekanAnni, line 100:
```suggestion
    ...mapActions(['buildMicroReactURL']),
```

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+      <button
+        v-if='!microreactToken'
+        @click='showModal()'
+        class='btn btn-block btn-standard btn-download'
+      >
+        Generate Microreact URL
+      </button>
+
+      <button
+        v-if='microreactToken && !results.perCluster[cluster]?.microreactURL'
```
LekanAnni, line 13: You could simplify this by adding the code in computed method, so you will only need to call the computed method here instead of writing the whole line. It encourages readability.

app/client/src/store/actions.ts
```diff
@@ -98,4 +98,45 @@ export default {
...
+    const { state } = context;
+    await api(context)
+      .withSuccess('addMicroreactURL')
+      .withError('addError')
+      .post<ClusterInfo>(`${config.server_url}/microreactURL`, {
+        cluster,
+        projectHash: state.projectHash,
+        apiToken: state.microreactToken,
+      });
+  },
```
LekanAnni, line 141:
```suggestion
    async buildMicroReactURL(
    context: ActionContext<RootState, RootState>,
    cluster: string | number,
    token: string
  ) {
    const { state, commit } = context;
    commit('setToken', token);
    await api(context)
      .withSuccess('addMicroreactURL')
      .withError('addError')
      .post<ClusterInfo>(`${config.server_url}/microreactURL`, {
        cluster,
        projectHash: state.projectHash,
        apiToken: state.microreactToken,
      });
  },
```

LekanAnni (changes requested):
Looking good, 

Please note that I could not run the app, a problem with the script building from MacBook.

app/client/src/components/GenerateMicroreactURL.vue
LekanAnni: You could also create an Enum where WRONG_TOKEN="Wrong Token"

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+        target='_blank'
+        rel='noreferrer'
+      >
+        Visit Microreact URL
+      </a>
+    </div>
+
+    <Modal v-if='isModalVisible' @close='closeModal' class='modalFlex'>
+      <template v-if='tokenWasWrong' v-slot:header>
+        Your submitted token seems to be wrong
```
Nicolas-Dolan, line 33:
```suggestion
        Your submitted token is invalid
```

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -0,0 +1,123 @@
...
+      </template>
+      <template v-else v-slot:header> No Token submitted yet </template>
+
+      <template v-if='tokenWasWrong' v-slot:body>
+        <p>It seems that the Token you gave us was wrong.</p>
+        <p>
+          Please make sure you are using a correct Token.<br />
+          You can find your token in your
```
Nicolas-Dolan, line 41: Seems strange to capitalise token. Either way it's best to be consistent with the capitalisation.

Nicolas-Dolan (commented):
A few minor comments, but the analysis wouldn't run for me so I couldn't do a full review.
image

LekanAnni (approved):
LGTM 👍 

Thanks for making changes...

richfitz (approved):

PR #16, opened 2022-09-23
Display network graph 

muppi1993:
Get graphml data from backend and display as graph.
Added buttons to switch between different clusters.

Currently, unit testing the CytoscapeGraph component does not work. When trying to mount the component to get a wrapper for the tests, I receive an Error:
```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TypeError: Cannot read properties of null (reading 'className')".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

For now I renamed that testfile to `CytoscapeGraph.spec_not_working_yet.ts` to not run this on CI, but this leaves the component untested.


app/client/src/cytoscape-graphml.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'cytoscape-graphml';
```
EmmaLRussell, line 1: I don't think you should need this here, as well as declaring in `shims-vue.d.ts`?

app/client/src/store/mutations.ts
```diff
@@ -50,11 +50,19 @@ export default {
...
     state.results.perCluster[URLinfo.cluster] = {
+      ...state.results.perCluster[URLinfo.cluster],
       cluster: URLinfo.cluster,
       microreactURL: URLinfo.url,
     };
   },
   setToken(state: RootState, token: string | null) {
     state.microreactToken = token;
   },
+  addGraphml(state: RootState, GraphInfo: Record<string, string>) {
```
EmmaLRussell, line 61:
```suggestion
  addGraphml(state: RootState, graphInfo: Record<string, string>) {
```

app/client/src/store/mutations.ts
```diff
@@ -50,11 +50,19 @@ export default {
...
+    state.results.perCluster[GraphInfo.cluster] = {
+      ...state.results.perCluster[GraphInfo.cluster],
```
EmmaLRussell, line 63: Why do you need the spread operator here? What might be retained from any previous graphInfo for the cluster?

EmmaLRussell (commented):
Looks great! The zoomability on the graph is really nice. 

Should there be some of heading on the graph clusters, or were these going to go onto a separate tab? I'm not sure about using buttons to change the displayed clusters, since buttons are typically used to do an action rathter than swith view. Could be link style or tabs? 

Have you committed your CytoscapeGraph unit test file? I can't see it in the changeset. I suspect the error might be related to `cytoscape` or `cytoscape-graphml' not dealing well with being loaded during unit testing rather than in a real browser context. You might need to mock one or both of these modules, as we typically do with e.g. plotly. Example [here](https://github.com/mrc-ide/wodin/blob/main/app/static/tests/unit/components/wodinOdePlot.test.ts#L1)

app/client/src/store/mutations.ts
muppi1993: I did this to prevent that Microreact URLs which get also stored in `state.results.perCluster[clusternumber]` are overwritten/ deleted. Also did the same the other way round in `addMicroreactURL()`, to prevent the graph data being deleted when already stored.

app/client/src/components/NetworkVisualisations.vue
```diff
@@ -1,44 +1,66 @@
...
   props: ['firstCluster'],
   data() {
     return {
+      visitedTabs: [this.firstCluster],
       selectedCluster: this.firstCluster,
     };
   },
   components: {
     CytoscapeGraph,
+    // BTabs,
```
EmmaLRussell, line 50: can remove these comment lines?

app/client/tests/unit/components/NetworkVisualisations.spec.ts
```diff
@@ -36,18 +36,18 @@ describe('Component displays one button per unique cluster and one network graph
     expect(wrapper.exists()).toBe(true);
   });
 
-  test('has all buttons and a graph element with class column90', () => {
-    expect(wrapper.findAll('.columnButtons')).toHaveLength(4);
-    expect(wrapper.findAll('.column90')).toHaveLength(1);
-    const graphComponent = wrapper.findAll('.column90')[0];
+  test('has all buttons and a graph element with class cytoscape-graph', () => {
```
EmmaLRussell, line 39: test name out of date now?

EmmaLRussell (approved):
Thanks, this looks great - I like the ui changes. 

For future reference, is there still a current mockup set that we're working towards? With the final layout in terms of top level tabs etc?

app/client/tests/unit/components/CytoscapeGraph.spec.ts
```diff
@@ -0,0 +1,125 @@
+// Mock the import of cytoscape
+/* eslint-disable import/first */
+
+const mockReady = jest.fn();
+const mockGraphMl = jest.fn();
+const mockCytoscape = jest.fn().mockReturnValue({
+  ready: mockReady,
+  graphml: mockGraphMl,
```
LekanAnni, line 8: Hmmmm!

This bit was the devil advocate with `Cytoscape` failing to mount.

LekanAnni (approved):
LGTM 👍

PR #17, opened 2022-09-30
Send filehash/name mapping with poppunk request

muppi1993:
This mapping is needed to replace the file hases in the result files with the right filenames.

To do: 

- [x] add filenames to project hash generation to avoid overwriting filenames from another user when file content is the same, but filename differs

### DIFF
--- a/app/client/src/components/GenerateMicroreactURL.vue
+++ b/app/client/src/components/GenerateMicroreactURL.vue
@@ -11,7 +11,7 @@
 
       <button
         v-if='tokenAvailable'
-        @click='buildMicroreactURL(cluster)'
+        @click='buildMicroreactURL({cluster, token: microreactToken})'
         class='btn btn-block btn-standard btn-download'
       >
         Generate Microreact URL
--- a/app/client/src/store/actions.ts
+++ b/app/client/src/store/actions.ts
@@ -45,16 +45,24 @@ export default {
   },
   async runPoppunk(context: ActionContext<RootState, RootState>) {
     const { state, commit } = context;
-    const jsonSketches = {} as { [key: string]: Record<string, never> };
-    // join all file hashes to create a project hash
-    const phash = Md5.hashStr(Object.keys(state.results.perIsolate).sort().join());
+    // generate filenameMapping that the backend can use to replace
+    // filehashes with filenames in results.
+    // Also generate a string of ordered hashes and corresponding filenames
+    // to be used to generate a projecthash that is unique to this combination
+    // of file contents and filenames
+    const filenameMapping = {} as { [key: string]: string | undefined };
+    let mappingOrdered = '';
+    Object.keys(state.results.perIsolate).sort().forEach((filehash) => {
+      filenameMapping[filehash] = state.results.perIsolate[filehash].filename;
+      mappingOrdered += filehash;
+      mappingOrdered += state.results.perIsolate[filehash].filename;
+    });
+    const phash = Md5.hashStr(mappingOrdered);
     commit('setProjectHash', phash);
+    // add all sketches to object
+    const jsonSketches = {} as { [key: string]: Record<string, never> };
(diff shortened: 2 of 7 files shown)


app/client/src/store/actions.ts
```diff
@@ -45,17 +45,33 @@ export default {
...
     const { state, commit } = context;
-    const jsonSketches = {} as { [key: string]: Record<string, never> };
-    // join all file hashes to create a project hash
-    const phash = Md5.hashStr(Object.keys(state.results.perIsolate).sort().join());
+    // generate filenameMapping that the backend can use to replace
+    // filehashes with filenames in results.
+    // Also generate a string of ordered hashes and corresponding filenames
+    // to be used to generate a projecthash that is unique to this combination
+    // of file contents and filenames
+    const filenameMapping = {} as { [key: string]: string | undefined };
```
EmmaLRussell, line 53:
```suggestion
    const filenameMapping = {} as Record<string, string | undefined>;
```
A little more readable maybe? Or we often define a `Dict` type as it's generally useful: `export type Dict<V> = { [k: string]: V }`  Could consider doing likewise?
Why would a value in the mapping ever be undefined?

app/client/src/store/actions.ts
```diff
@@ -45,17 +45,33 @@ export default {
...
+    let mappingOrdered = '';
+    Object.keys(state.results.perIsolate).sort().forEach((filehash) => {
+      filenameMapping[filehash] = state.results.perIsolate[filehash].filename;
+      mappingOrdered += filehash;
+      mappingOrdered += state.results.perIsolate[filehash].filename;
+    });
+    const phash = Md5.hashStr(mappingOrdered);
     commit('setProjectHash', phash);
+    // add all sketches to object
+    const jsonSketches = {} as { [key: string]: Record<string, never> };
```
EmmaLRussell, line 63: Why `never`? Is there a not a real type for these? I'm not really sure why this line has changed, is it related to the file name mapping?

app/client/src/components/GenerateMicroreactURL.vue
```diff
@@ -11,7 +11,7 @@
 
       <button
         v-if='tokenAvailable'
-        @click='buildMicroreactURL(cluster)'
+        @click='buildMicroreactURL({cluster, token: microreactToken})'
```
EmmaLRussell, line 14: Not sure how this change is related to the filename mapping?

EmmaLRussell (commented):
If I understand it, this branch just allows sending filenames along with filehashes in the request so that the results which come back (using the related backend branch) include real file names rather than hashes. When I run the analysis I now see the filenames next to the nodes on the graph, so I guess that means it's worked!(?)

app/client/src/store/actions.ts
muppi1993: when removing undefined, line 56 would give an typescript error for `filenameMapping[filehash]`: `Type 'string | undefined' is not assignable to type 'string'`. I guess it is undefined for the empty filenameMapping without any elements in it?

app/client/src/components/GenerateMicroreactURL.vue
muppi1993: It is not.. Just realised that the behaviour of generating the microreactURL was sometimes not as expected and found this little bug, where I had forgotten to provide all parameters to the action. Thought this fix is so small I can just add that with this PR rather than opening a new PR just for this one line of code?

app/client/src/store/actions.ts
muppi1993: Yes, this should always have a value, since the action can only be triggered when files have been uploaded. However, adding '!' causes `warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion` on linting, shall I just switch that warning off?

EmmaLRussell (approved):

PR #18, opened 2022-10-13
example test mocking cytoscape

EmmaLRussell:
I had a go at testing that `ready` had been invoked, and that it does what we expect it to. Quite tricky because of there being so many different 3rd party methods being called. 

Also, I think having an arrangement where the wrapper is generated outside any test call, and also having multiple `describe` blocks, makes managing the mocks a little harder. I usually just have one `describe` per module, and create all wrappers in `getWrapper` methods.


muppi1993 (approved):
This is great, thank you so much Emma!

PR #19, opened 2022-10-17
Highlight query samples in network graph 

muppi1993:
Border-colour of query samples should be red to highlight own samples in the network graph.


EmmaLRussell (approved):
So the query clusters are the ones from the files the user has uploaded as opposed to the ones that were already in the db?
LGTM. I wonder if it would be good to have a short explanation or legend to this effect? Thought that could be in another ticket (or is there another epic around help/docs anyway?)

muppi1993:
Yes, that's right. I think we can add the legend in a future PR, since the plan is to add more metadata like serotype or AMR status in the future and allow to change the nodes colours based on these metrics. This will require a legend, and then we could add explanation on the red circles as well.

PR #20, opened 2022-10-20
bacpop-66 make config fully flexible across environments

hillalex:
The change to the front-end here is that instead of reading in a json settings file, settings are provided by a js file, so that individual settings can be functions. This allows the same settings file to apply to any deployed environment, regardless of url, because the settings can read from `window.location` to determine the deployed location. This is the same approach used in [montagu-webapps](https://github.com/vimc/montagu-webapps).

The change to the back-end is that config is read in using `fs` when the app is started, rather than needing to exist at compile time. This means the docker image can contain the compiled code and the config can still be injected in after the container is brought up (but before the app runs).


EmmaLRussell (commented):
Nice. I wondered about maybe putting in an interface to the settings stuff in the front end, so that only one source file is reading from `@settings` so that approach could be swapped out if need be - but I don't think it would be worth the complexity of adding an extra layer for that. 

Should probably add an explanation of all this to the README.

EmmaLRussell (approved):
Thanks!

PR #21, opened 2022-10-25
bacpop-65 changes to proxy for constellation

hillalex:
Based off [bacpop-66](https://github.com/bacpop/beebop/pull/20) - a few tweaks to allow the proxy to play nice with the new [beebop-deploy](https://github.com/bacpop/beebop-deploy/pull/1)


EmmaLRussell (commented):
So this is essentially allowing the name of the server container to be configured via the deployment constellation, rather than fixed in the proxy template to `beebop-server`?

hillalex:
> So this is essentially allowing the name of the server container to be configured via the deployment constellation, rather than fixed in the proxy template to `beebop-server`?

yeah exactly, make sure it matches the name of the constellation deployed container rather than hard-coded

EmmaLRussell (approved):

PR #24, opened 2022-12-02
bacpop-76 Save user-project hash mapping to redis when run poppunk

EmmaLRussell:
This branch adds saving user id to project hash mapping whenever a request to run poppunk is receieved. The project hash is included in the poppunk request, and the saved user id is constructed from both the auth provider and user id available on the request. 

A Redis class and UserStore class have been added, following a pattern already used in Wodin, as well as an `asyncHandler` which ensures that the express server does not crash during async request handling, by wrapping server code, and calling `next` on error. 

There is not yet an endpoint to retrieve user project data, but an integration test has been added to check that the expected value is saved to redis. This means that we do need to have redis running while running the integration tests, so I've added running of components to the backend 'unit and integration tests' pipeline/ 

Package lock churn, sorry!

### DIFF
--- a/app/server/tests/integration/utils.ts
+++ b/app/server/tests/integration/utils.ts
@@ -1,0 +1,39 @@
+import axios from "axios";
+import Redis from "ioredis";
+
+const fullUrl = (url: string) => `http://localhost:4000/${url}`;
+const redisUrl = "redis://localhost:6379";
+
+// Do not fail on error status - let individual tests deal with error cases
+const validateStatus = () => true;
+const standardHeaders = { "Content-Type": "application/json" };
+
+export const get = async (url: string) => {
+    return axios.get(fullUrl(url), { headers: standardHeaders, validateStatus });
+};
+
+export const post = (url: string, payload: any, cookie: string) => {
+    const headers = { ...standardHeaders, "Cookie": cookie };
+    return axios.post(fullUrl(url), payload, {headers, validateStatus });
+};
+
+const withRedis = async (func: (redis: Redis) => any) => {
+    const redis = new Redis(redisUrl);
+    try {
+        return await func(redis);
+    } finally {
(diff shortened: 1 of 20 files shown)


scripts/run_test
```diff
@@ -13,7 +13,7 @@ docker run --rm -v $VOLUME:/beebop/storage mrcide/beebop-py:$API_BRANCH \
        ./scripts/download_db --small storage
 docker network create $NETWORK > /dev/null || /bin/true
 
-docker run -d --rm --name $NAME_REDIS --network=$NETWORK redis:5.0
+docker run -d --rm --name $NAME_REDIS --network=$NETWORK -p 6379:6379 redis:5.0
```
EmmaLRussell, line 16: Exposing the redis port here so we can access it from the app when running outside docker.

app/server/tests/integration/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// Do not fail on error status - let individual tests deal with error cases
+const validateStatus = () => true;
+const standardHeaders = { "Content-Type": "application/json" };
+
+export const get = async (url: string) => {
+    return axios.get(fullUrl(url), { headers: standardHeaders, validateStatus });
+};
+
+export const post = (url: string, payload: any, cookie: string) => {
```
richfitz, line 15: do you want to add an "allow explicit any" here? Don't we normally have something to allow this in tests generally?

richfitz (approved):
This all looks nice and familiar.

Purely from an avoiding churn point of view it looks like something is different in your editor config here with lots of missing trailing newlines?

app/server/tests/integration/utils.ts
EmmaLRussell: I think that should be sorted now

PR #26, opened 2023-01-04
bacpop-77 Create project with name

EmmaLRussell:
This PR requires users to enter a name when creating a new project, and persists that name to redis when the analysis is started. Some related UI changes are also included, Future branches will provide management and reloading of a user's previous projects. 

Key changes:
- the client includes the project name along with the project hash in the request to run poppunk
- however this means that the request format to the beebop client now differs from the poppunk request format (that request fails when projectName is included) so there are now two server request types, and the server strips out projectName when forwarding the request to poppunk
- the projectName is saved to redis, associated with both the projectHash and the user (since different users could create projects with the same hash if they used multiple files, but may name them differently). 
- front end UI changes to start converting the Home page into a Projects page as shown in [mockup](https://docs.google.com/presentation/d/1htgjPhrGWn8R7aYiJ3A4DEEUUOv6A7rlXB_NhEJrf7s/edit?usp=sharing). Removed unimplemented buttons and home icon, and added input and button for creating project with a required name. 
- Added redirect from the analysis page back to the home page if project has no name - require user to provide a name before proceeding. 
- Removed the welcome user message in the home page, but added a 'Logged in as' label to the header so that info is always visible. 
- Project name is shown as header in the analysis page. NB there is a separate epic for restyling the analysis page more fully

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ The backend can be tested with
 ```
 npm run test
 ```
-insinde `app/server`.
+from `app/server`.
 
 ### End-to-end tests
 To run end-to-end test, the app must be started with `./scripts/run_test`. In a new terminal, these test can be launched with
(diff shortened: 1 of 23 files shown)


LekanAnni (approved):
LGTM 👍 

I still can't bring the app up. Is there a documentation for this app ?

PR #27, opened 2023-01-09
bacpop-95 eslint rule changes

EmmaLRussell:
This branch adds our standard set of front end eslint rules to beebop's client package, in order to be consistent with our other applications. It uses an identical `eslintrc.js` to that used in wodin. The code has been updated to conform to these rules. The main changes are:
- indentation of 4 spaces rather than 2
- double quotes for string literals rather than single
- max line length of 120
- no dangling commas after the final prop in objects

Sorry, it's a big changeset, but all the changes are simple!

### DIFF
--- a/app/client/.editorconfig
+++ b/app/client/.editorconfig
@@ -1,6 +1,6 @@
 [*.{js,jsx,ts,tsx,vue}]
 indent_style = space
-indent_size = 2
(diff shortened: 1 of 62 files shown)


LekanAnni (approved):
Looks great 👍

PR #29, opened 2023-02-07
bacpop-97 Clear store state on new project

EmmaLRussell:
This branch clears the store state on new project so that a new project does not load with the files etc from any previous loaded project. Will be used in conjunction with [bacpop-98](https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-98) which should ensure latest project state is always persisted. 

Introduces `emptyState()` util, which is used to initialise the first store state, and to create fresh new projects. Note that this will also clear non-project store props, like versions and savedProjects - but these are always refreshed from API when needed, so it shouldn't matter..


richfitz (approved):

M-Kusumgar (approved):
Looks pretty straight forward! Nicely done

PR #34, opened 2023-06-16
bacpop-83 Load project status

EmmaLRussell:
This branch ensures that loaded project progress status data is injected into the store state, and also that polling is restarted if the project was not finished when it was previously unloaded. We also needed to ensure that polling is stopped when the Project View is unloaded and user browses back to the Home page.

When the user loads a finished project, all data should be available to view and download. When the user loads a project which is still in progress, it should complete successfully and all data should be available when it does. 

In the course of doing this, it made sense to refactor 'submittedStatus' to be a boolean since that was essentially how it was behaving, and I also renamed it to 'submitted'.

Should be merged along with the corresponding beebop_py branch which it targets. 

Load errors are occasionally encountered if load is attempted before cluster allocation has completed - there's a ticket to fix that [here](https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-131/Allow-project-to-load-before-cluster-allocation-has-completed)


M-Kusumgar (commented):
Apologies for the delay! Was trying to bulldoze through the hint stuff. The code itself looks good, don't have any complaints but struggling to actually test it. Am I right in thinking that I need to upload a fasta file for the project to get a hash? I do end up getting a hash as the payload of the "status" request however, the app seems to be polling forever? Is this normal because of how slow it is? Perhaps would be better to have a call and have a run through of this branch. Also a sample FASTA file would be great! Might be one already but wasn't obvious to me, I have some example ones which seem to take forever!

Also given how much time it seems to take, perhaps 1000ms as the polling interval is a little quick haha

EmmaLRussell:
> Apologies for the delay! Was trying to bulldoze through the hint stuff. The code itself looks good, don't have any complaints but struggling to actually test it. Am I right in thinking that I need to upload a fasta file for the project to get a hash? I do end up getting a hash as the payload of the "status" request however, the app seems to be polling forever? Is this normal because of how slow it is? Perhaps would be better to have a call and have a run through of this branch. Also a sample FASTA file would be great! Might be one already but wasn't obvious to me, I have some example ones which seem to take forever!
> 
> Also given how much time it seems to take, perhaps 1000ms as the polling interval is a little quick haha

Sorry, thought you already had access to some fasta examples - I've attached a zip of these to the KB article: https://mrc-ide.myjetbrains.com/youtrack/articles/bacpop-A-1/beebop

Not sure why you were seeing the issues with polling forever, though I was having some issues with poppunk last week (or thought I was..). Let's maybe have a look at it later today or tomorrow.

scripts/run_dependencies
```diff
@@ -2,7 +2,7 @@ NETWORK=beebop_nw
 VOLUME=beebop-storage
 NAME_REDIS=beebop-redis
 NAME_API=beebop-py-api
-API_BRANCH=main
+API_BRANCH=bacpop-83
```
richfitz, line 5: don't forget to change this once https://github.com/bacpop/beebop_py/pull/28 is merged

richfitz (approved):

PR #35, opened 2023-06-27
bacpop-79 add timestamp to saved project

EmmaLRussell:
Saves, returns and displays the date when a project was created. For display, just uses the standard locale date string, which should be appropriate.

Also, updates the application title and favicon from the default.


app/client/public/favicon.ico
richfitz, line 1: nice :)

app/client/tests/e2e/LoggedIn.spec.ts
```diff
@@ -111,7 +111,10 @@ test.describe("Logged in Tests", () => {
         await expect(page.locator("#cy canvas")).toHaveCount(3);
         // can browse back to Home page and see new project in history
         await page.goto(config.clientUrl());
-        await expect(await page.locator(".saved-project-row").last()).toHaveText("test project", { timeout });
+        await expect(await page.locator(".saved-project-row .saved-project-name").last())
+            .toHaveText("test project", { timeout });
+        expect(await (await page.locator(".saved-project-row .saved-project-date").last()).innerText())
+            .toMatch(/^[0-3][1-9]\/[0-1][1-9]\/20[2-9][0-9], [0-2][0-9]:[0-5][0-9]:[0-5][0-9]$/);
```
richfitz, line 117:
```suggestion
            .toMatch(/^[0-3][0-9]\/[0-1][0-9]\/20[2-9][0-9], [0-2][0-9]:[0-5][0-9]:[0-5][0-9]$/);
```

otherwise 30-10-2022 won't match. Date regex is hard (impossible even)

richfitz (changes requested):
LGTM, but I think your test regex is wrong

app/client/tests/e2e/LoggedIn.spec.ts
EmmaLRussell: oops!

M-Kusumgar (commented):
Looks good! Yh just realised Rich is right about the date regex, following the discussion about removing seconds from time and possibly removing the comma, here are some useful links!

[removing seconds](https://www.w3schools.com/jsref/jsref_tolocalestring.asp)
[removing comma](https://stackoverflow.com/questions/49982572/how-to-remove-comma-between-date-and-time-on-tolocalestring-in-js)

EmmaLRussell:
As discussed with @M-Kusumgar I've tersified the date format a bit (no seconds, no comma between date and time) - this meant using a different strategy for unit testing, so it's now mocking the `Date.toLocaleString` method. This was being flaky in CI for node 12, so I've taken the opportunity to cull the node version matrix being tested - not really sure why we were doing that anyway!

M-Kusumgar (approved):
Looks good to me!

richfitz (approved):

PR #36, opened 2023-06-27
bacpop-132 Handle loading projects where sample cluster have not yet been assigned

EmmaLRussell:
This branch, along with its [associated branch in beebop_py](https://github.com/bacpop/beebop_py/pull/30), which is targets, should allow projects to be loaded without error before clusters have been assigned (this is the first thing that happens when you run an analysis and typically happens pretty quickly, within the first few seconds). 

To test this you'll want to add a reasonable number of fasta files to a project, so that there's a bit of a delay between submitting the job for analysis and the clusters being assigned. It works for me with around 10 files, but note that I've seen the front end processing get stuck with a project using all 17 example fasta files ([ticket to fix this](https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-134))

Hit the "Start Analysis" button then quickly browse to the Home page, then click on the project name in the list. The project should successfully load with clusters not yet assigned to the samples, but the cluster numbers should be loaded shortly afterwards. Note that sketch data will also not be visible before the clusters are assigned, because these are saved out together when the cluster job finishes. We could deal with this by saving out sketches first, but not sure it would be worth it. 

Contains changes from https://github.com/bacpop/beebop/pull/34

### DIFF
--- a/app/client/src/store/mutations.ts
+++ b/app/client/src/store/mutations.ts
@@ -90,8 +90,8 @@ export default {
         state.analysisStatus = projectResponse.status;
         // TODO: We can currently assume that if we've got a successful response then the project had been submitted as
         // you get an error if it hadn't because beebop_py won't know about it. When bacpop-98 is done, we should get
-        // indication of submitted status in the response for all projects including those which had never been submitted
-        // and will need to update this line accordingly.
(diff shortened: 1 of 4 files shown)


EmmaLRussell:
Closing this are changes are no longer required on the beebop side.

PR #37, opened 2023-07-13
bacpop-137 Refactor routes module into separate routes and controllers modules

EmmaLRussell:
This branch splits up the monolithic routes module, into separate `routes` and `controllers` for `index` (top level routest), `project` and `user` endpoints. The routes modules defines the endpoints available, and the controllers provide the implementations previously present in the `apiEndpoints` object. 

This is just to make the routes a bit more modular so the files are of manageable length.

Also added a few missing types.

Lots of line changes but just a re-org of existing logic.


app/server/docker/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM node:16
 
 COPY ./package.json /app/package.json
 WORKDIR /app
-RUN npm install
+RUN npm ci
```
EmmaLRussell, line 5: Match the CI workflow

richfitz (approved):

PR #39, opened 2023-07-14
bacpop-80 Display number of samples in project list

EmmaLRussell:
Includes sample count in displayed project information on home page. 

Had some trouble with the Playwright tests, so I've added a step in the workflow to upload failing test screenshots as an artefact. 

I've also tried to rationalise some of the timeouts in the Playwright tests to avoid excessively long failing test runs by:
- removing fixed timeouts while waiting for processes to run, and replacing them with a wait for element values, which will instead check continuously until the expectation is true
- cutting down the default timeout for most operations, and using a custom timeout of 3 minutes for the longest wait, for the analysis to run

I also updated the checks on browse back to project page to check the first project in the list, as these are now in reverse chronological order. 

To further reduce test speed locally, I've also changed it to only test against one browser when not run within CI.

### DIFF
--- a/app/client/playwright.config.ts
+++ b/app/client/playwright.config.ts
@@ -21,7 +21,7 @@ const firefox = {
 const config: PlaywrightTestConfig = {
     testDir: "./tests/e2e",
     /* Maximum time one test can run for. */
-    timeout: 1000 * 60 * 3,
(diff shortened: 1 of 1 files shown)


M-Kusumgar (approved):
Looks good to me!

richfitz (approved):

PR #40, opened 2023-08-02
bacpop-138 Validate project name change

EmmaLRussell:
This branch adds some front end validation when user edits project name - change will not be allowed to name which is empty or whitespace, or if another of the user's projects has the same name. The same rules will be added to Create Project in a separate ticket. 

Key changes:
- added a getter to do the actual check, as we'll also want to use this for creating new projects and an enum for the possible check results. 
- updated `EditProjectName` component to check candidate project name on each key change, prevent saving when validation fails and show an appropriate message (implemented as a component, which will also be used in the Create Project case). To avoid user alarm, there is no error message or button disable if the name is identical to the current name, but no action will be invoked if click 'Save' at this point. 
- Removed 'Project' prefix from Project page header - I don't think it was useful, and it was causing formatting headaches when editing!
- Includes SavedProjects when state is reset on Project load, so that they are available for checking duplicate names

Adding the no duplicates requirement has impacted the e2e tests - I have modified where required, and will eventually need to update others, once create project prevents duplicates too..

### DIFF
--- a/app/client/src/components/projects/EditProjectName.vue
+++ b/app/client/src/components/projects/EditProjectName.vue
@@ -1,5 +1,5 @@
 <template>
-  <span v-if="!editingProjectName">
+  <template v-if="!editingProjectName">
         <slot></slot>
         <i class="bi bi-pencil edit-icon clickable"
            v-b-tooltip.hover
@@ -7,8 +7,8 @@
            @click="editProjectName"
            v-on:keyup.enter="editProjectName"
         ></i>
-    </span>
-    <span v-else>
+    </template>
+    <template v-else>
         <input ref="renameProject"
                class="project-name-input"
                type="text"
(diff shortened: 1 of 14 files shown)


M-Kusumgar (changes requested):
Looks good, however I think we should only allow one edit at a time, maybe turn off edit mode on blur, because if you do the steps below, you can name two projects the same thing
1. Create two projects with whatever names
2. Toggle both to edit mode on the projects home page
3. Change them to the same name and save

EmmaLRussell:
> Looks good, however I think we should only allow one edit at a time, maybe turn off edit mode on blur, because if you do the steps below, you can name two projects the same thing
> 
>     1. Create two projects with whatever names
> 
>     2. Toggle both to edit mode on the projects home page
> 
>     3. Change them to the same name and save

Ha, yep, good point!

EmmaLRussell:
Made a few little changes to support only allowing one project name at a time to be edited:
- cancel the edit when the component loses focus - added an enclosing div to catch this event
- make sure that the component actually has focus when editing starts! Focus the input when edit button is first pressed - do this on nextTick to ensure the input is available.
- Set a fixed height on the component which is the same whether in editing or non-editing mode - this prevents UI jumping around and also avoids problem of click event on the icon getting lost because the focus out on an above component jumped up the UI so the click location was no longer over the icon.

M-Kusumgar (approved):
Just tested it out! Looks good and the code looks good! The way I did it in wodin is just by having it go to non-edit mode on blur and the handleBlur function also checks if event.relatedTarget was the button or not, so that is one way you can customise blur events so that they will let through events from certain other objects in the UI! But mousedown works too!

PR #41, opened 2023-09-26
bacpop-110 Persist Microreact tokens

EmmaLRussell:
This branch persists microreact token when user enters one, sending it to the back end to be saved in the redis db, encrypted using the `crypto` package. 

When the user logs in, any token is fetched, and this will be used next time they generate a Microreact url, rather than requiring them to re-enter their token after each log in. 

This means the user doesn't currently have any way to delete or change their token - this will be done in a separate ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-141

The github actions are running against the "in.development" config, which uses various values from vault - but they don't ever actually attempt to get those values from the vault (using the `decrypt_config` script).. and I'm not sure we can use the vault in gha anyway? Hence using the hardcoded fake encrypt key (which is integration tested in the gha workflow).. but I think that means we've probably got some vault values which never actually get used as they aren't the "PROD" ones, e.g. `GOOGLE_ID`, `GOOGLE_SECRET` (and must not actually get used in any of the integration tests).


EmmaLRussell:
Closing as we'll implement in front end instead

PR #43, opened 2023-10-13
bacpop-144 Fix docker image and docker testing

EmmaLRussell:
I noticed that the beebop docker image was broken - this branch fixes a few issues, updates node to the latest version and adds a smoke test for running dockerised beebop  so we'll know right away next time!

Most of the line count diff is package lock changes. 

I'd like to fix up the docker script so that if run locally it supports auth - currently this isn't the case so you can't properly manually test beebop using `run_docker` - but it sounds like this would be really useful for reviewing since it's such a pain to set up locally! But one for another ticket...

### DIFF
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -23,7 +23,7 @@
     "cytoscape": "^3.23.0",
     "cytoscape-graphml": "^1.0.6",
     "jquery": "^3.6.1",
-    "node-sass": "^7.0.1",
+    "node-sass": "^9.0.0",
     "sass-loader": "^12.6.0",
     "ts-md5": "^1.2.11",
     "ts-node": "^10.7.0",
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -24,7 +24,7 @@
     "cytoscape-graphml": "^1.0.6",
     "jquery": "^3.6.1",
     "node-sass": "^9.0.0",
-    "sass-loader": "^12.6.0",
+    "sass-loader": "^13.3.2",
     "ts-md5": "^1.2.11",
     "ts-node": "^10.7.0",
     "vue": "^3.2.13",
--- a/app/server/tsconfig.json
+++ b/app/server/tsconfig.json
@@ -1,9 +1,10 @@
 {
     "compilerOptions": {
-        "target": "es2019",
+        "target": "esnext",
+        "module": "commonjs",
         "esModuleInterop": true,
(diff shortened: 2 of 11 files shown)


app/client/package.json
```diff
@@ -23,8 +23,8 @@
     "cytoscape": "^3.23.0",
     "cytoscape-graphml": "^1.0.6",
     "jquery": "^3.6.1",
-    "node-sass": "^7.0.1",
-    "sass-loader": "^12.6.0",
+    "node-sass": "^9.0.0",
+    "sass-loader": "^13.3.2",
```
EmmaLRussell, line 27: Packages which required later version for node 20

app/server/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
     "compilerOptions": {
-        "target": "es2019",
+        "target": "esnext",
+        "module": "nodenext",
```
EmmaLRussell, line 4: Specify latest es and module support - this allows both `import`s and `require`s, which seemed to be the main problem in the docker container. I'm not sure why this had stopped working for docker, but ran ok on metal. :shrug:

scripts/run_docker
```diff
@@ -12,7 +12,7 @@ else
 fi
 export HOST
 docker-compose up -d
-docker cp app/server/src/resources/config.json beebop_beebop-server_1:/app/src/resources/config.json
+docker cp app/server/src/resources/config.json.in.docker beebop_beebop-server_1:/app/src/resources/config.json
```
EmmaLRussell, line 15: We were just using the wrong config file before in this script, which mean the redis location was wrong.

app/server/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
     "compilerOptions": {
-        "target": "es2019",
+        "target": "esnext",
+        "module": "commonjs",
         "esModuleInterop": true,
         "resolveJsonModule": true,
         "outDir": "dist",
-        "moduleResolution": "Node"
+        "moduleResolution": "node"
```
EmmaLRussell, line 8: This combination of values allows both `import`s and `require`s in the server code. I'm not sure why this had stopped working in the container but not on metal... There are also options for `nodenext` which should work too, but get compilation complaints about instability and demands to use TS nightly builds..

richfitz (approved):

M-Kusumgar (approved):

PR #50, opened 2024-02-26
Bacpop 156 - Initial unit tests

absternator:
This PR contains the setup of unit testing and unit tests the current ui-rewrite branch.

### DIFF
--- a/app/client-v2/e2e/vue.spec.ts
+++ b/app/client-v2/e2e/vue.spec.ts
@@ -1,8 +1,0 @@
-import { test, expect } from '@playwright/test';
-
-// See here how to get started:
-// https://playwright.dev/docs/intro
-test('visits the app root url', async ({ page }) => {
-  await page.goto('/');
-  await expect(page.locator('div.greetings > h1')).toHaveText('You did it!');
-})
--- a/app/client-v2/src/__tests__/composables/useTheme.spec.ts
+++ b/app/client-v2/src/__tests__/composables/useTheme.spec.ts
@@ -1,0 +1,67 @@
+import { DARK_THEME, LIGHT_THEME, useTheme } from "@/composables/useTheme";
+import { usePrimeVue } from "primevue/config";
+import { ref } from "vue";
+import { useStorage } from "@vueuse/core";
+
+// vitest.mock("@vueuse/core", () => ({
+//   useStorage: vitest.fn().mockReturnValue(ref(LIGHT_THEME))
+// }));
+const mockChangeTheme = vitest.fn();
+vitest.mock("primevue/config", () => ({
+  // replace with actual import path
+  usePrimeVue: vitest.fn(() => ({
+    changeTheme: mockChangeTheme
+  }))
+}));
+
+describe("useTheme", () => {
+  it("should toggle theme from light to dark", () => {
+    const { toggleTheme, themeState } = useTheme();
+
+    toggleTheme();
+
+    expect(themeState.value).toBe(DARK_THEME);
+    expect(mockChangeTheme).toHaveBeenCalledWith(LIGHT_THEME, DARK_THEME, "theme-link", expect.any(Function));
+  });
+
+  it("should toggle theme from dark to light", () => {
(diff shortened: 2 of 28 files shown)


EmmaLRussell:
I thought you were going to break this up into sensible sized PRs?

app/client-v2/src/config.ts
```diff
@@ -1,7 +1,9 @@
 export const getApiUrl = () => {
   if (import.meta.env.MODE === "development") {
     return "http://localhost:4000";
-  } else {
-    return "https://${window.location.host}/api";
   }
+  if (import.meta.env.MODE === "test") {
+    return "";
+  }
```
EmmaLRussell, line 7: I thought most of the tests for urls were for `stringContaining` so do you need this? Or is it for the mock server part?

app/client-v2/src/layouts/AppNav.vue
```diff
@@ -52,7 +52,7 @@ const toggle = (event: MouseEvent) => {
       </template>
       <template #end>
         <div class="flex gap-3">
-          <div><Button :icon="themeIcon" @click="toggleTheme" outlined /></div>
+          <div><Button aria-label="theme-switcher" :icon="themeIcon" @click="toggleTheme" outlined /></div>
```
EmmaLRussell, line 55: `aria-label` values should be human-consumable shouldn't they? Typically read out by a screen reader, so punctuation maybe not an issue, but I'd expect this to be formatted like a label. 
```suggestion
          <div><Button aria-label="Switch theme" :icon="themeIcon" @click="toggleTheme" outlined /></div>
```

app/client-v2/src/views/HomeView.vue
```diff
@@ -114,7 +115,7 @@ const onRowEditSave = async (event: DataTableRowEditSaveEvent) => {
             <div>
               <InputGroup>
                 <InputText placeholder="Create new Project" v-model="newProjectName" @keydown.enter="addProject" />
-                <Button icon="pi pi-plus-circle" @click="addProject" />
+                <Button aria-label="add-project" icon="pi pi-plus-circle" @click="addProject" />
```
EmmaLRussell, line 118:
```suggestion
                <Button aria-label="New project" icon="pi pi-plus-circle" @click="addProject" />
```

EmmaLRussell (changes requested):
Great stuff!

The tests aren't actually running on gha because the branch isn't targeting main. I had the same issue recently. see [this comment](https://github.com/bacpop/beebop-deploy/pull/4#pullrequestreview-1895420406) for Rich's suggested approach. 

Also, on BuildKite, the image isn't being pushed because of:
```
Error parsing reference: "mrcide/beebop-server:bacpop-156/initial-unit-test" is not a valid repository/tag: invalid reference format
```
I think that's probably because of the slash in the branch name.

Need to modify codecov too I guess to make sure we get coverage for the new front end tests.

app/client-v2/e2e/vue.spec.ts
```diff
@@ -1,8 +0,0 @@
-import { test, expect } from '@playwright/test';
```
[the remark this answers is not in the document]
absternator: was just the boiler plate one from create vue

app/client-v2/src/__tests__/composables/useTheme.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { DARK_THEME, LIGHT_THEME, useTheme } from "@/composables/useTheme";
+import { usePrimeVue } from "primevue/config";
+import { ref } from "vue";
+import { useStorage } from "@vueuse/core";
+
+// vitest.mock("@vueuse/core", () => ({
+//   useStorage: vitest.fn().mockReturnValue(ref(LIGHT_THEME))
+// }));
```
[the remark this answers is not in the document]
absternator: good catch :smile:

app/client-v2/src/config.ts
absternator: yeah i did but just changed to make things simpler to match in general.. so future tests and stuff

PR #53, opened 2024-03-01
Bacpop-156-initial-tests

absternator:


app/client-v2/src/layouts/AppNav.vue
```diff
@@ -61,6 +61,7 @@ const toggle = (event: MouseEvent) => {
               @click="toggle"
               aria-haspopup="true"
               aria-controls="overlay_menu"
+              aria-label="user-menu"
```
EmmaLRussell, line 64:
```suggestion
              aria-label="User Menu
```

EmmaLRussell (commented):
Looks good! :sunglasses: 
Still need to add code coverage I think. 
Smoke test seems to be failing, not sure why.

EmmaLRussell (approved):
Nice!

PR #55, opened 2024-03-05
Bacpop-157 project page pt1

absternator:
Member
[absternator](https://github.com/absternator) commented [yesterday](https://github.com/bacpop/beebop/pull/54#issue-2166581360)
Includes the first part of project page. This has everything done up to the point where a person runs analysis.
Note: everything in the public folder and all the wasm and worker logic is copy pasted over from current FE... Thus most logic is same as current fe

There will 2 following prs that will complete the functionality of the project.


app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const baseApi = mande(getApiUrl(), { credentials: "include" });
+
+// TODO: add proper error handling. Maybe best to add error state attribute and watch accordingly cos of nested things interval/workers
+export const useProjectStore = defineStore("project", {
+  state: () => ({
+    basicInfo: {} as Pick<Project, "id" | "name" | "timestamp">,
+    fileSamples: [] as ProjectSample[],
+    projectHash: "",
+    isRun: false,
```
EmmaLRussell, line 23: Does this mean has started running, or has completed running?  Maybe `started` would be better if the former.

app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
...
+  type WorkerResponse,
+  WorkerResponseValueTypes
+} from "@/types/projectTypes";
+import { mande } from "mande";
+import { defineStore } from "pinia";
+import { Md5 } from "ts-md5";
+
+const baseApi = mande(getApiUrl(), { credentials: "include" });
+
+// TODO: add proper error handling. Maybe best to add error state attribute and watch accordingly cos of nested things interval/workers
```
EmmaLRussell, line 17: Got a ticket for this? Might as well add it to the epic so we we don't forget.

app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
+import { getApiUrl } from "@/config";
```
EmmaLRussell, line 1: This file isn't hugely readable - I wouldn't be averse to splitting it up into state, getters and actions - or at least putting in a bit of whitespace between sections and methods,

app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
...
+import { mande } from "mande";
+import { defineStore } from "pinia";
+import { Md5 } from "ts-md5";
+
+const baseApi = mande(getApiUrl(), { credentials: "include" });
+
+// TODO: add proper error handling. Maybe best to add error state attribute and watch accordingly cos of nested things interval/workers
+export const useProjectStore = defineStore("project", {
+  state: () => ({
+    basicInfo: {} as Pick<Project, "id" | "name" | "timestamp">,
```
EmmaLRussell, line 20: Why this Pick? Why isn't the state here just of type Project? What's that type used for if not this?

app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
...
+import {
+  COMPLETE_STATUS_TYPES,
+  type AnalysisStatus,
+  type ApiResponse,
+  type Project,
+  type ProjectSample,
+  type WorkerResponse,
+  WorkerResponseValueTypes
+} from "@/types/projectTypes";
+import { mande } from "mande";
```
EmmaLRussell, line 11: Should we be using `mande` in the other places which are doing API access?

app/client-v2/src/stores/projectStore.ts
```diff
@@ -0,0 +1,132 @@
...
+          name: projectRes.data.name,
+          timestamp: projectRes.data.timestamp
+        };
+        this.fileSamples = projectRes.data.samples;
+        this.projectHash = projectRes.data.hash || "";
+
+        if (projectRes.data.status) {
+          this.isRun = true;
+          this.analysisStatus = projectRes.data.status;
+        }
```
EmmaLRussell, line 61: I feel like this would be neater if it was constructing a Project object (as suggested above) and then setting it atomically in the state.

app/client-v2/src/__tests__/components/ProjectView/NotRunProject.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  it("should render drag and drop section no files uploaded", () => {
+    render(NotRunProjectVue, {
+      global: {
+        plugins: [createTestingPinia(), PrimeVue]
+      }
+    });
+
+    expect(screen.getByText(/drag and drop/i)).toBeVisible();
+  });
+  it("should call store.runanalysis on run analysis click and not disabled", async () => {
```
EmmaLRussell, line 19:
```suggestion
  it("should call store.runAnalysis on run analysis click and not disabled", async () => {
```

EmmaLRussell (commented):
Nice, looks really clean! :smiley: 

As discussed, drop area seems to be a bit flaky, fails unpredictably on both single and multiple files, with no errors logged. Maybe make a ticket to investigate that?

I wonder if users will find it a bit jarring that they can only drop initial files and subsequently have to use file dialog - it's an inconsistency that it would be nice to avoid since the old app does allow them to drag additional files. Maybe the drop area could slim down and sit under the Upload button? I also wonder if it's going to be obvious to users what the Upload button is for when they're already looking at a big area for uploading files. I've seen components before where you've got a big drop area and the button and its text are incorporated into it, so "Drag and drop files here to upload, or [select files]" (e.g. https://vaadin.com/directory/component/drop-or-choose-upload). And then potentially that whole thing could slim down and sit on the left once files are added. 

A still think the new project control should be on the left of the Home screen - I keep typing in the search box when I'm trying to make a new project!  :rofl: 

Various subjective gripes about naming!

I think there's a fair bit in here related to running projects which have been copied over from the old front end around running projects, polling, checking if complete etc, which I'm assuming are going to be cleaned up and used in the next ticket.

Do we have a ticket to add browser tests?

app/client-v2/src/components/ProjectView/NotRunProject.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
```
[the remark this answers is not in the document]
absternator: have upated to ProjectPreRun and ProjectPostRun

absternator:
> Nice, looks really clean! 😃
> 
> As discussed, drop area seems to be a bit flaky, fails unpredictably on both single and multiple files, with no errors logged. Maybe make a ticket to investigate that?
> 
> I wonder if users will find it a bit jarring that they can only drop initial files and subsequently have to use file dialog - it's an inconsistency that it would be nice to avoid since the old app does allow them to drag additional files. Maybe the drop area could slim down and sit under the Upload button? I also wonder if it's going to be obvious to users what the Upload button is for when they're already looking at a big area for uploading files. I've seen components before where you've got a big drop area and the button and its text are incorporated into it, so "Drag and drop files here to upload, or [select files]" (e.g. https://vaadin.com/directory/component/drop-or-choose-upload). And then potentially that whole thing could slim down and sit on the left once files are added.
> 
> A still think the new project control should be on the left of the Home screen - I keep typing in the search box when I'm trying to make a new project! 🤣
> 
> Various subjective gripes about naming!
> 
> I think there's a fair bit in here related to running projects which have been copied over from the old front end around running projects, polling, checking if complete etc, which I'm assuming are going to be cleaned up and used in the next ticket.
> 
> Do we have a ticket to add browser tests?

1. Hvae created tickets browser tests, and yes the project page pt2 PR will contain all the relevant work.... 

2. With regard to the file upload have made a ticket to investigate its intermency. THe other comment though we can still drag and drop when files have already been uploaded .. the goal was to mimic google drive....

app/client-v2/src/components/ProjectView/NotRunProject.vue
```diff
@@ -0,0 +1,49 @@
...
+import { useProjectStore } from "@/stores/projectStore";
+import ProjectDataTable from "@/components/ProjectView/ProjectDataTable.vue";
+
+const store = useProjectStore();
+</script>
+
+<template>
+  <FileUpload
+    name="genomics[]"
```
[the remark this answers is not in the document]
absternator: removed

app/client-v2/src/components/ProjectView/NotRunProject.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
+import { useProjectStore } from "@/stores/projectStore";
+import ProjectDataTable from "@/components/ProjectView/ProjectDataTable.vue";
+
+const store = useProjectStore();
+</script>
+
+<template>
```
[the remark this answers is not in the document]
absternator: as per ending comment we are trying to mimic google drive... where you can still drag and drop on the table

app/client-v2/src/components/ProjectView/AmrColumn.vue
```diff
@@ -0,0 +1,61 @@
...
+const props = defineProps<{
+  amr?: AMR;
+}>();
+console.log(props.amr);
+
+const tagSeverity = (percent: number) => {
+  if (percent < 33) return "danger";
+  if (percent < 66) return "warning";
+  if (percent > 66) return "success";
+  return "secondary";
```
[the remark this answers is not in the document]
absternator: yup good idea can double check with him

app/client-v2/src/components/ProjectView/AmrColumn.vue
```diff
@@ -0,0 +1,61 @@
+<script setup lang="ts">
+import type { AMR } from "@/types/projectTypes";
+import { computed } from "vue";
+
+const props = defineProps<{
+  amr?: AMR;
+}>();
+console.log(props.amr);
```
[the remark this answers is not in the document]
absternator: good catch haaha

app/client-v2/src/components/ProjectView/AmrColumn.vue
```diff
@@ -0,0 +1,61 @@
...
+};
+const getPercent = (value: number) => Math.round(value * 100);
+
+const displayAMR = computed(() => {
+  if (!props.amr) return;
+  return {
+    Penicillin: {
+      label: "P",
+      value: getPercent(props.amr.Penicillin),
+      severity: tagSeverity(getPercent(props.amr.Penicillin))
```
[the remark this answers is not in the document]
absternator: great suggestion applied

app/client-v2/src/stores/projectStore.ts
absternator: good idea made ticket

app/client-v2/src/stores/projectStore.ts
absternator: yup if we have another store we can if we don't need to inside a component... vueuse useFetch is available inside components however

app/client-v2/src/stores/projectStore.ts
absternator: good idea il do all this project store comments in the pt2 PR as it will have to restructure in that as well

app/client-v2/src/stores/projectStore.ts
absternator: good idea il do all this project store comments in the pt2 PR as it will have to restructure in that as well

app/client-v2/src/stores/projectStore.ts
absternator: good idea il do all this project store comments in the pt2 PR as it will have to restructure in that as well

EmmaLRussell:
>  THe other comment though we can still drag and drop when files have already been uploaded .. the goal was to mimic google drive...

Oh really, what if you drag onto the header area? Hmm, I don't think that's obvious enough, because the 'Drop here' label disappears, right?

app/client-v2/src/stores/projectStore.ts
EmmaLRussell: Why not do it here and merge into pt2? Shouldn't be too many conflicts unless you've done major restructuring between pt1 and 2, which seems unlikely. This is partly me being selfish as it will be easier for me to review this way though! 😆

absternator: sorry i just already did it :cry: ... the merge would've just been nasty to do because of restructuring

absternator:
> > THe other comment though we can still drag and drop when files have already been uploaded .. the goal was to mimic google drive...
> 
> Oh really, what if you drag onto the header/table area? Hmm, I don't think that's obvious enough, because the 'Drop here' label disappears, right? Maybe just needs an extra little bit of text to tell people that they can do this.

Updated to add label on hover as well for clarrity
image

app/client-v2/src/stores/projectStore.ts
EmmaLRussell: Ah no worries

EmmaLRussell (approved):
> Updated to add label on hover as well for clarrity

That looks good, but I don't think it really solves the problem as it only appears when you drag over and the issue is that the uninformed user may not realise you *can* still drag over, as there isn't any indication on screen (as there was before the first file was uploaded). I think if that new label was visible all the time, that would do the trick though. 

Looks good other than that, happy to merge this if it helps things along.

PR #56, opened 2024-03-05
Bacpop-158 project pg pt2

absternator:
This contains the part 2 of the project page which handles the running state and the state after the project has been run.
Note: most logic is coppied over from the exisitng FE. For dev testing it should work as expected as per exisiting FE.

The next part of this project page will contain the microreact url and network graphs work.


absternator:
Address comments abouts `projectStore.ts` from PR #55

app/client-v2/src/components/ProjectView/MicroReactColumn.vue
```diff
@@ -0,0 +1,32 @@
...
+import { type ProjectSample, AnalysisType } from "@/types/projectTypes";
+
+const store = useProjectStore();
+const props = defineProps<{
+  data: ProjectSample;
+}>();
+</script>
+
+<template>
+  <div v-if="store.project.status?.microreact === 'finished'" class="flex gap-2">
```
EmmaLRussell, line 12: I wonder if we shold have an enum for statuses like "finished" rather than using magic strings..

app/client-v2/src/stores/projectStore.ts
```diff
@@ -112,21 +107,133 @@ export const useProjectStore = defineStore("project", {
...
+    stopPollingStatus() {
+      if (this.pollingIntervalId) {
+        clearInterval(this.pollingIntervalId);
+        this.pollingIntervalId = null;
+      }
     },
     // TODO: update to remove from api as well
     removeUploadedFile(index: number) {
-      this.fileSamples.splice(index, 1);
+      // this.fileSamples.splice(index, 1);
```
EmmaLRussell, line 175: Is there a separate ticket to implement file remove?

app/client-v2/src/stores/projectStore.ts
```diff
@@ -112,21 +107,133 @@ export const useProjectStore = defineStore("project", {
...
+
     async runAnalysis() {
-      console.log("Run");
-      this.isRun = true;
+      const body = this.buildRunAnalysisPostBody();
+      try {
+        await baseApi.post("/poppunk", body);
+        this.project.hash = body.projectHash;
+      } catch (error) {
+        console.error("Error running analysis", error);
```
EmmaLRussell, line 184: I think I've already asked this, but there is a ticket for putting in error handling generally?

EmmaLRussell (commented):
Looks great! Didn't pay too much attention to the run logic - as you say, it looks like it's mostly been pulled over from the old client. 

I think the flashing 'Running Analysis' text is a bit distracting. Maybe if it flashed a bit slower, or was replaced with animated ellipsis, it would be less full on..?

This is from the last PR, but I'm just wondering if actually the colours for the AMR should be reversed. A high number means high likelihood of resistance to the drug, so it feels like that ought to have the danger colour rather than the "Everything's fine!" colour...

One for another ticket, but it would be nice to have some indication in the home page if a project is pre-run, running or completed. Can't remember if you had something like that in the mockups?

app/client-v2/src/components/ProjectView/MicroReactColumn.vue
absternator: i have used union types for this statusus already :smile:

app/client-v2/src/stores/projectStore.ts
absternator: yup

absternator:
> Looks great! Didn't pay too much attention to the run logic - as you say, it looks like it's mostly been pulled over from the old client.
> 
> I think the flashing 'Running Analysis' text is a bit distracting. Maybe if it flashed a bit slower, or was replaced with animated ellipsis, it would be less full on..?
> 
> This is from the last PR, but I'm just wondering if actually the colours for the AMR should be reversed. A high number means high likelihood of resistance to the drug, so it feels like that ought to have the danger colour rather than the "Everything's fine!" colour...
> 
> One for another ticket, but it would be nice to have some indication in the home page if a project is pre-run, running or completed. Can't remember if you had something like that in the mockups?

1. have slowed down animation 
2. have swapped colors around 
3. Yeah I can make ticket.. tried to do in FE only but needs backend work to return something to know a project is finished.. can currently do one if has been run like image
image

image

app/client-v2/src/components/ProjectView/MicroReactColumn.vue
EmmaLRussell: Yeah ok, less nice but more efficient!

EmmaLRussell (approved):

PR #63, opened 2024-03-21
Bacpop-141 Update Microreact token and delete microreact

absternator:
This PR allows users to update and delete their incorrect token... The dialog has been extracted into its component so it can be reused.

Initially, we wanted to put this functionality in the user menu, but the issue is we need a project context and have it be fully run  to be able to check If the token is valid or not... thus have put icon with tooltip in microreact column.

Testing: Ensure can update and delete microreact token. Only Saving when the token is correct when updating.

image
image

### DIFF
--- a/app/client-v2/components.d.ts
+++ b/app/client-v2/components.d.ts
@@ -19,6 +19,7 @@ declare module 'vue' {
     Menu: typeof import('primevue/menu')['default']
     Menubar: typeof import('primevue/menubar')['default']
     MicroReactColumn: typeof import('./src/components/ProjectView/MicroReactColumn.vue')['default']
+    MicroReactTokenDialog: typeof import('./src/components/ProjectView/MicroReactTokenDialog.vue')['default']
     NetworkGraph: typeof import('./src/components/ProjectView/NetworkGraph.vue')['default']
     NetworkTab: typeof import('./src/components/ProjectView/NetworkTab.vue')['default']
     ProgressBar: typeof import('primevue/progressbar')['default']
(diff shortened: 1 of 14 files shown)


EmmaLRussell (changes requested):
UI looks great. It does seems a bit weird to not be able to change this at the user level though - maybe there's another way we could test the token than generating the url? But I think this is fine for now. 

Could you put a tooltip on the cog icon when it's disabled as otherwise it's a bit confusing. Something like "Microreact settings will become available when Microreact data has been generated". 

When I delete my token from the Update or Delete dialog,  after clicking the Delete button I'm taken back out to the Save Microreact Token dialog - it should just close the first dialog.

There's also some weird behaviour around putting in an invalid token. I'd assumed that if I did this, it would still retain the previous token, which it doesn't. That's probably ok, but it doesn't close the dialog if I click Delete subsequently (consistent with the behaviour above), and it continues to show the error message (so I have to cancel out of the dialog to escape from it). Probably Delete should be removed or disabled at this point if in fact the token has been cleared behind the scenes? Also, the old error message is still displayed when I re-open the dialog after close.

absternator:
> Could you put a tooltip on the cog icon when it's disabled as otherwise it's a bit confusing. Something like "Microreact settings will become available when Microreact data has been generated".
> 
> When I delete my token from the Update or Delete dialog, after clicking the Delete button I'm taken back out to the Save Microreact Token dialog - it should just close the first dialog.
> 
> There's also some weird behaviour around putting in an invalid token. I'd assumed that if I did this, it would still retain the previous token, which it doesn't. That's probably ok, but it doesn't close the dialog if I click Delete subsequently (consistent with the behaviour above), and it continues to show the error message (so I have to cancel out of the dialog to escape from it). Probably Delete should be removed or disabled at this point if in fact the token has been cleared behind the scenes? Also, the old error message is still displayed when I re-open the dialog after close.

1. Yeah thats what i thought so i spent time looking into the microreact api but found nothing... the issue is the microreact api just returns a internal server error when the token is wrong not a 401 or helpfull message 
2. Have added tooltip on disabled and made dialog close on delete 
3. Since i made dialog close on delete i think last one should be okay now. Also made it so error goes away on re-open dialog like suggested

EmmaLRussell (approved):
Great, thanks!

PR #64, opened 2024-03-22
Bacpop-159 V8 DB updates 

absternator:


EmmaLRussell (approved):
Yeah, thanks this is great - when you revert API_BRANCH back to main, could you keep `run_docker` as it is now, i.e. using API_BRANCH rather than hard-coded to main?

PR #68, opened 2024-04-10
Add 'delete project' button

david-mears-2:
https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-86/User-can-delete-a-project

Requesting eagle eyes for this review, because it uses several technologies I'm not familiar with:
- Express
- Redis
- Typescript

TODO before 'ready to merge':
- [x] Handle requests on deleted projects e.g. get, renaming, deleting samples, etc


absternator (commented):
Awesome work looks really nice and clean code !!! Just left a few comments... And love all the cleanups !!!!

Sorry whats the todo before merge? didnt understand it


Ps. will test soon just waiting for popubk db to download

david-mears-2:
> Sorry whats the todo before merge? didnt understand it

I checked it off when I submitted the commit [Add validation checks in userStore for deleted projects](https://github.com/bacpop/beebop/pull/68/commits/f88911fccdff9aa342b5020aac2ca19dcc560a0a) - prevents actions being performed on deleted records and things associated with them.

app/client-v2/src/views/HomeView.vue
```diff
@@ -48,6 +51,38 @@ const addProject = async () => {
   router.push(`/project/${project.value?.data}`);
 };
 
+const deleteProject = async (index: number) => {
```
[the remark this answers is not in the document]
david-mears-2: Yeah nice idea

app/client-v2/src/types/projectTypes.ts
```diff
@@ -41,6 +41,7 @@ export interface Project {
     microreact: StatusTypes;
     network: StatusTypes;
   };
+  deleted_at?: string;
```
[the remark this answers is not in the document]
david-mears-2: My guidance from Rich and Emma was to use a flag - was discussed in [Slack](https://resideglobal.slack.com/archives/C05467EB6Q2/p1712323158551949).

app/client-v2/setupTests.ts
```diff
@@ -2,7 +2,7 @@ import { server } from "@/mocks/server";
 import "@testing-library/jest-dom/vitest";
 
 beforeAll(() => {
-  server.listen();
+  server.listen({ onUnhandledRequest: "error" });
```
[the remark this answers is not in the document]
david-mears-2: Maybe better to have another place they might be caught just in case? That's my vote, will wait and see how reviewer 2 votes.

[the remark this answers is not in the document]
david-mears-2: Possibly, someone will introduce something later that doesn't (yet) use a default handler. I can see how it's a bit duplicatey, though.

app/server/src/db/userStore.ts
```diff
@@ -61,64 +81,75 @@ export class UserStore {
...
     }
     async deleteSample(projectId: string, sampleHash: string, filename: string) {
+        await this._validProject(projectId);
         const sampleId = this._sampleId(sampleHash, filename);
         await this._redis.srem(this._projectSamplesKey(projectId), sampleId);
         await this._redis.del(this._projectSampleKey(projectId, sampleId));
     }
 
     async getSample(projectId: string, sampleHash: string, filename: string): Promise<{ sketch: Record<string, unknown>; amr: AMR }> {
+        await this._validProject(projectId);
```
david-mears-2, line 136: This gets repetitive/cluttery. Does anyone know if there is a way to specify "run this util before each function in the following list" at the top of the file or something?

app/client-v2/src/views/HomeView.vue
```diff
@@ -48,6 +51,38 @@ const addProject = async () => {
...
+
+  confirm.require({
+    message: `Are you sure you want to delete the project '${projects.value?.data[index].name}'?`,
+    header: projects.value?.data[index].name,
+    icon: "pi pi-exclamation-triangle",
+    rejectLabel: "Cancel",
+    acceptLabel: "Delete project",
+    rejectClass: "p-button-secondary p-button-outlined",
+    acceptClass: "p-button-danger",
+    reject: () => {
```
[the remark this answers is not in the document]
absternator: well i guess the project wont be gone from the list right? maybe its just me I've never seen an info popup show up on cancel

app/client-v2/src/types/projectTypes.ts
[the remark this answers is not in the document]
absternator: OO also just noticed- namng convention is camelCase in JS.. so should be `deletedAt`

app/server/src/db/userStore.ts
absternator: yeah thats what i was thinking too but nothing nice honestly I could think of

app/client-v2/src/views/HomeView.vue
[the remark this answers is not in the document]
EmmaLRussell: Yeah, I don't think you'd typically need toast on cancel.

app/server/src/db/userStore.ts
```diff
@@ -21,9 +21,19 @@ export class UserStore {
...
 
     private async _setProjectName (projectId: string, newProjectName: string) {
+        await this._validProject(projectId);
         await this._redis.hset(this._projectKey(projectId), "name", newProjectName);
     }
 
+    private async _validProject(projectId: string): Promise<Record<string, unknown>> {
+        const project = await this._redis.hgetall(this._projectKey(projectId));
+        if (project.deleted_at) {
+            throw new Error(`This project has been deleted.`);
```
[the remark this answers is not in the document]
EmmaLRussell: I'm surprised lint didn't moan about that, possibly a missing rule.

app/client-v2/setupTests.ts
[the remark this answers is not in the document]
EmmaLRussell: Yeah, the default handler will in fact handle everything and return an empty success response so `onUnhandledRequest` will never be triggered, so there isn't a lot of point to this change.

I suppose there's an argument to had about whether it would be better to omit the default handler and require the tests to explicitly handle all expected requests. But the handlers are quite disconnected from the individual tests so they're not really part of the testing of expected behaviour, it's just potentially slightly confusing when odd behaviour happens in the test because you forgot the code was expecting a response from some endpoint. But that's a separate issue!

app/server/src/db/userStore.ts
EmmaLRussell: Maybe you could use set up separate middleware that runs before the main route handler, which checks if there is a projectId in request parameters and if so validates it?

app/client-v2/src/__tests__/views/HomeView.spec.ts
```diff
@@ -198,4 +199,116 @@ describe("HomeView ", () => {
...
+          expect(screen.getByText(/Are you sure you want to delete the project/i)).toBeVisible();
+        });
+
+        const confirmDeleteButton = await screen.getByRole("button", { name: "Delete project" });
+        await userEvent.click(confirmDeleteButton);
+
+        expect(fetchSpy).toHaveBeenCalledWith(
+          `${projectIndexUri}/${MOCK_PROJECTS[0].id}/delete`,
+          expect.objectContaining({ credentials: "include" })
+        );
```
EmmaLRussell, line 235: This seems like a nice pattern!

app/client-v2/src/__tests__/views/HomeView.spec.ts
```diff
@@ -198,4 +199,116 @@ describe("HomeView ", () => {
...
+        await screen.findByRole("cell", {
+          name: MOCK_PROJECTS[0].name
+        });
+
+        const firstDeleteButton = await screen.getByRole("button", {
+          name: new RegExp(`delete ${MOCK_PROJECTS[0].name}`, "i")
+        });
+        await userEvent.click(firstDeleteButton);
+
+        // Mock a different response to 'get' (and, later, assert that this different set of projects is listed) to verify that get is called a second time
```
EmmaLRussell, line 218: Yeah, good idea. I guess an alternative would be to use fetchSpy again. But this way you also check that it gets rendered correctly.

app/server/src/db/userStore.ts
EmmaLRussell, line 31: If you throw a `BeebopError` here instead, you can include a message and status code that will make its way back to the client in the response.

README.md
```diff
@@ -142,9 +142,16 @@ npm run test:dev
...
+```
+npx playwright install-deps
+```
+
+In a new terminal, these tests can be launched with
 ```
 npx playwright test
 ```
-from `app/client/`.
+from `app/client-v2/`.
```
EmmaLRussell, line 156: Yeah, thanks for cleaning this up!

EmmaLRussell (commented):
Great stuff, works like a charm!

I think if the user attempts to browse back to a deleted project's page, it should ideally display an error message reflecting the reason, rather than a generic message as currently. The server is actually throwing a 500. Should probably be a 404, and message should say that the project does not exist, or has been deleted. 

It would also be good to include backend integration test for deleting a project, and for triggering the error when you attempt to do something with a project which has been deleted.

app/server/src/routes/projectRoutes.ts
```diff
@@ -15,6 +15,9 @@ export default {
         app.get('/project/:projectId',
             authCheck,
             controller.getProject);
+        app.patch('/project/:projectId/delete',
```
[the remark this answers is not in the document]
david-mears-2: It seems we are sending a request body with the patch request that deletes project samples. Delete does not normally have a body, so I guess that's why it uses patch.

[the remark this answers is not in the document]
EmmaLRussell: Yeah that's true, and also because it's not a straightforward delete of a resource. So that should stay as patch, but project delete should indeed be DELETE I think.

app/server/src/db/userStore.ts
david-mears-2: I haven't done this, let me know if you think it's blocking

app/client-v2/src/components/ProjectView/ProjectPage.vue
```diff
@@ -3,12 +3,17 @@ import { useProjectStore } from "@/stores/projectStore";
...
+import { ref, onMounted, onUnmounted } from "vue";
 import Toast from "primevue/toast";
 
 const route = useRoute();
 const store = useProjectStore();
-const projectFetchError = await store.getProject(route.params.id as string);
+
+let projectFetchError = ref<any>(null);
+
+onMounted(async () => {
```
absternator, line 14: sorry why is this being done on mount? we ideally want to fetch data before anything has rendered

app/client-v2/src/components/ProjectView/ProjectPage.vue
```diff
@@ -36,7 +47,7 @@ onUnmounted(() => {
     </div>
     <div class="surface-card p-4 shadow-2 border-round">
       <ProjectPostRun v-if="store.startedRun" />
-      <ProjectPreRun v-else />
+      <ProjectPreRun v-if="store.project && store.project.samples" />
```
absternator, line 50: also why the change here? we either want to see postrun or pre run

david-mears-2: These changes were to try and get around some error I was getting - but now if I revert the changes, the tests all still pass, and the page still works. So I'm not sure what was going on, but we don't need these changes.

app/client-v2/src/components/ProjectView/ProjectPage.vue
david-mears-2: See below.

app/client-v2/src/components/ProjectView/ProjectPage.vue
david-mears-2: Reverted!

absternator (approved):
LGTM! nicueee

app/server/src/db/userStore.ts
EmmaLRussell: No, just a thought!

EmmaLRussell (approved):
Great stuff, thanks!

Just a note on review process - if you're the reviewee, please avoid resolving conversations initiated by a reviewer's comments - the reviewer is going to need to open those back up anyway in order to check them when they re-review. If you just want to keep track of which comments you've addressed, you could add your own comment or emoji instead. The exception is when you accept suggestions and github does the commit and resolves the convo automatically... Sorry, should have mentioned this!

PR #69, opened 2024-06-18
Bacpop-172 - feat: amr column catogory + color updates

absternator:
This PR updates the AMR column display.. It now is reverted similar to old style where color gradient changes dependent on resistance and also on tooltip hover shows the probability expressed as word (e.g likely, highly likely)... The code for choosing colours and word is has the same functionality as before but just cleaned up.

new: 
image
old: 
image

old before UI rewrite: 
image


absternator:
[fasta_examples.zip](https://github.com/user-attachments/files/15901738/fasta_examples.zip)

app/client-v2/src/utils/amrDisplayUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    return "Unsure";
+  }
+  const thresholds = antibioticProbabilityMap[antibiotic];
+  for (const { threshold, word } of thresholds) {
+    if (probability >= threshold) {
+      return word;
+    }
+  }
+  return "Unsure";
+};
```
M-Kusumgar, line 72: "Unsure" actually means something and has a numerical value (for `Chloramphenicol`) so we should not be displaying that if probability is not a number or if it wasnt greater than a threshold, I would say add in a key called "Undefined" and maybe have the transparency 0 for that key and text bit more greyed out to give it a disabled look, I tried a random fasta file from the internet and got something that visibly looks like it works even though probability was `NaN`

M-Kusumgar (commented):
Looks good, love the color, was definitely worth having a think about! Got one comment but happy to approve after that!

app/client-v2/src/utils/amrDisplayUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (typeof probability !== "number") {
+    return "Unknown";
+  }
+  const thresholds = antibioticProbabilityMap[antibiotic];
+  for (const { threshold, word } of thresholds) {
+    if (probability >= threshold) {
+      return word;
+    }
+  }
+  return "Unsure";
```
M-Kusumgar, line 73: If it has made it through all the other things then probably unknown as well right?
```suggestion
  return "Unknown";
```

M-Kusumgar (approved):
LGTM, small suggestion but yh happy to approve!

PR #72, opened 2024-07-01
Bacpop-173 e2e tests

absternator:
These are some basic e2e tests which cover the basic functionality of the app


app/client-v2/e2e/home.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  await expect(page.getByText("Project name already exists or is empty")).toBeVisible();
+});
+
+test("can not add duplicate project name or empty name", async ({ page }) => {
+  await page.getByPlaceholder("Create new Project").fill("");
+  await page.getByPlaceholder("Create new Project").press("Enter");
+  await expect(page.getByText("Project name already exists or is empty")).toBeVisible();
+});
+
+test("can navigate delete project", async ({ page }) => {
```
M-Kusumgar, line 32: i think
```suggestion
test("can delete project", async ({ page }) => {
```

app/client-v2/playwright.config.ts
```diff
@@ -43,26 +43,33 @@ export default defineConfig({
     headless: !!process.env.CI
   },
 
-  /* Configure projects for major browsers */
+  /* Configure projects for chrome. Slow running tests thus not worth running on all browsers*/
```
M-Kusumgar, line 46: hmmm i think there is value in running these tests on all browsers, often times these tests dont only catch actual bugs (usually unit/integration tests are good at catching them) but they also can catch when youve broken stuff on safari or something cos who uses safari :laughing: 

if speed is an issue then i think making these run in parallel should do the trick

M-Kusumgar (commented):
There is a testing failing on CI but also have some comments about the not running on different browsers, i think it is worth doing and the speed shouldnt be a problem if you do a different github action for each browser for example

Also the projectPostRun file has 4 slow tests, even if you parallelise this, github runners only have 2 cores so at most you can only half, another way to do it is to split this test file up even more and run those individual files in parallel too

app/client-v2/playwright.config.ts
EmmaLRussell: Could potentially do an all-browsers tests nightly rather than on every push if worried about speediness - but we'd need to be sure we'd notice the failure!

M-Kusumgar (approved):

PR #75, opened 2024-07-15
Bacpop-175- Fix Upload Samples 

absternator:
This PR solves the problem of uploading too many samples at one time. The PR includes the following changes: 
- BE: add `addSamples` endpoint that allows array of samples with amr and sketch. (before only could do 1 amr or sketch)
- Main logic in `projectstore.ts - processFiles` function. The upload including amr and sketch is done in batches of 5 and limits the amount of web workers to half the users machine. This ensures computer can handle and does not run out of memory as was happening previously.
- Web worker - now handles arrays of files and posts message after fully processing all array of files 
- Removed sketch column - not needed as amr and sketch are done together and batched and added to list (performance reasons as above)
- Progress bar when uploading samples

Testing: 
uploading lots of samples at 1 time!!! if you need more please contact me :smile:


app/client-v2/src/stores/projectStore.ts
```diff
@@ -75,50 +76,76 @@ export const useProjectStore = defineStore("project", {
...
+        worker.onmessage = async (event: MessageEvent<WorkerResponse[]>) => {
+          this.handleWorkerResponse(event.data);
+          worker.terminate();
+          resolve();
         };
+
         worker.onerror = (error) => {
           console.error(error);
-          this.toast.showErrorToast("Ensure uploaded sample file is correct");
+          this.toast.showErrorToast("Ensure uploaded sample files are correct, or try again later.");
```
EmmaLRussell, line 135: In the case that one or more uploaded samples failed the sketching and AMR process, the user would see this message, but there would be no other indication of what samples had failed, other than that they would be missing from the list displayed - is that correct? I think it would be good for the user to see a list of files which failed in this case, as it's going to be tough for them to work it out if they have hundreds of files. 

Also, I think it's the case that if one sample in batch fails, the remaining successful samples from the batch will not be added the project (since the worker won't get as far as sending the message with the samples array) - I think it would be good if the worker could be more defensive and output both successful and failed samples so successful samples can be included. Again, could be in another ticket!

app/client-v2/src/stores/projectStore.ts
```diff
@@ -75,50 +76,76 @@ export const useProjectStore = defineStore("project", {
...
-              }
-        );
+        await baseApi.post(`/project/${this.project.id}/sample`, samples);
       } catch (error) {
         console.error(error);
-        this.toast.showErrorToast("Ensure uploaded sample file is correct or try again later.");
-        if (matchedHashIndex !== -1) {
-          this.project.samples.splice(matchedHashIndex, 1);
-        }
+        this.toast.showErrorToast("Ensure uploaded sample files is correct or try again later.");
```
EmmaLRussell, line 147:
```suggestion
        this.toast.showErrorToast("Ensure uploaded sample files are correct or try again later.");
```
..but also, we should revisit how useful this is in the light of possibly processing hundreds of samples. Again, it would be better if the user could get some record of which samples failed which part of the process.

app/server/src/controllers/projectController.ts
```diff
@@ -140,5 +140,21 @@ export default (config) => {
         sendSuccess(response, null);
       });
     },
+
+    async addSamples(request, response, next) {
```
EmmaLRussell, line 144: Can we also get rid of the old singular endpoint?

app/server/src/controllers/projectController.ts
```diff
@@ -140,5 +140,21 @@ export default (config) => {
...
+      await asyncHandler(next, async () => {
+        const samples = request.body as {
+          sketch: Record<string, unknown>;
+          hash: string;
+          amr: AMR;
+          filename: string
+        }[];
```
EmmaLRussell, line 151: Should really be a type for this in models.

app/server/src/db/userStore.ts
```diff
@@ -121,6 +121,27 @@ export class UserStore {
         await this._redis.hset(this._projectSampleKey(projectId, sampleId), "amr", JSON.stringify(amr));
     }
 
+    async saveSamples(projectId: string, samples: {
```
EmmaLRussell, line 124: Can we get rid of `saveSketch` and `saveAMR` now?

EmmaLRussell (commented):
This looks great, thanks, big improvement! Couple of suggestions, which could potentially go in separate tickets. 

I haven't tested it with hundreds of files - did you put the 700 sample set from Nick somewhere?

EmmaLRussell (commented):
Thanks for sending the full file set through, works great with 762 files!

That lag is definitely noticeable before the progress bar appears, and even before the 'Sketching and calculating' message appears, when it isn't clear if anything is happening - so yeah, would be good to get more early user feedback in. 

The paging of samples, 50 to a page, is good. I wonder if it would also be useful for these huge projects to have an indication of the number of samples in the project somewhere in the page, and also maybe the ability to search for samples in the list or on the network tab. We could maybe ask Nick about that. 

It was quite unresponsive when I tried to Run Analysis, though it did manage it eventually. If I then clicked away to project page and then attempted to reload, I saw a spinner with close tag (!)... but then it successfully loaded the running project. 
image
Quite a lot of the samples failed, was that expected?

app/client-v2/src/stores/projectStore.ts
```diff
@@ -75,50 +76,76 @@ export const useProjectStore = defineStore("project", {
...
+          const fileHash = Md5.hashStr(content);
+          return { hash: fileHash, filename: file.name, file };
+        })
+      );
+      const hashedFileBatches: HashedFile[][] = [];
+      for (let i = 0; i < hashedFiles.length; i += batchSize) {
+        hashedFileBatches.push(hashedFiles.slice(i, i + batchSize));
+      }
+      return hashedFileBatches;
+    },
```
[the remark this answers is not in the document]
absternator: yeah it gets bad too with lots of files!!! i tried to see if quick fix but couldn't find it.. thus have created this [ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-177/Fix-Uploading-initial-wait) where it will be fixed

app/client-v2/src/stores/projectStore.ts
absternator: this case ive found never really happens.. I've only noticed it when there is an intermittent error with my machine / web workers... thus I've just added a generic message.. but il make ticket regardless as your point is 100% correct

app/server/src/controllers/projectController.ts
absternator: i just left just incase we need it but will remove makes sense

absternator:
> a page, is good. I wonder if it would also be useful for these huge projects to have an indication of the number of samples in the project somewhere in the page, and also maybe the ability to search for samples in the list or on the network tab. We could maybe ask Nick about that.

>That lag is definitely noticeable before the progress bar appears, and even before the 'Sketching and calculating' message appears, when it isn't clear if anything is happening - so yeah, would be good to get more early user feedback in.
 
> It was quite unresponsive when I tried to Run Analysis, though it did manage it eventually. If I then clicked away to project page and then attempted to reload, I saw a spinner with close tag (!)... but then it successfully loaded the running project.

1. have created this [ticket](https://mrc-ide.myjetbrains.com/youtrack/agiles/103-46/current?issue=bacpop-179)
2. as per above comment its to do with file uploader component.. have made ticet to sort it 
3. yeah I noticed that too lots of failed... I'm actually not sure I looked at logs and they failed for valid reasons but unsure.... haha noticed closing tag and have removed :laughing:

EmmaLRussell (approved):
Great, thanks!

PR #76, opened 2024-07-17
feat: Update nginx.conf.template to serve all routes from the front-end Vue app

absternator:
this has been deployed to https://beebop-dev.dide.ic.ac.uk/.   to test head to website and create project, then refresh page and ensure still sees project and not 404


EmmaLRussell (approved):
Looks good - can we do without the /about section now we've got the catch all?

PR #78, opened 2024-09-04
Bacpop-181 CSV export project

absternator:
The following PR allows for users to export the project in csv format. The button shown in below screenshot. the csv will have columns filename, (...amr for each antibiotic), cluster.... if there is no cluster the text _run for cluster assignment_ will show.

Testing: 
1. create project and add fasta samples and run. Then click export button and csv should download that looks correct as per columns labelled above.

image

### DIFF
--- a/app/client-v2/e2e/projectPostRun.spec.ts
+++ b/app/client-v2/e2e/projectPostRun.spec.ts
@@ -89,3 +89,12 @@ test("can run project multiple times", async ({ page }) => {
   await expect(page.getByText("GPSC7")).toBeVisible();
   await expect(page.getByText("GPSC4")).toBeVisible();
 });
+
+test("can export project data as csv", async ({ page }) => {
+  uploadFiles(page);
+
+  const downloadPromise = page.waitForEvent("download");
+  await page.getByLabel("Export").click();
+  const download = await downloadPromise;
+  expect(download.suggestedFilename()).toBe(`${projectName}.csv`);
+});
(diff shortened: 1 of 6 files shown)


EmmaLRussell (commented):
Looks great - couple of tiny potential nitpicks:
- should maybe be consistent about capitalising the column headers - currently all the drug names are capitalised but cluster and filename are not,
- I think silently putting "run for cluster assignment" into the file seems a bit strange. I think it would be better to warn user "Cluster values will be blank in exported file, as analysis has not yet run. Export anyway?" - and get them to confirm, then just leave cluster values empty. 
- Commas aren't escaped in the csv generation, so you end up with an extra column in the row if you've got a filename with a comma in it (should wrap these values in double quotes I think). I tried this and found that that sample failed analysis anyway - probably because of a similar issue in the back end! But that showed up another issue with putting "run analysis.." if the cluster is missing, as it shows that for failed samples too, where running won't make a difference. I think having cluster be blank for those samples would be fine too.

EmmaLRussell:
> I can't entirely agree with this as users that use beebop know whats going on. so if they export without cluster assignment that should be fine... putting a dialog may be annoying and seems like its a bad thing they are doing (when its entirely valid).

I see what you mean, but it just seems weird to pollute output data with a user instruction, especially if we're assuming that they know what they're doing... Could include 'don't warn again' in the dialog, or could pop up some non-intrusive toast saying "Exported without cluster values (run analysis to include next time)" or something like that.

EmmaLRussell:
> users that use beebop know whats going on

Dangerout assumption! 😆

absternator:
> I see what you mean, but it just seems weird to pollute output data with a user instruction, especially if we're assuming that they know what they're doing... Could include 'don't warn again' in the dialog, or could pop up some non-intrusive toast saying "Exported without cluster values (run analysis to include next time)" or something like that.

yup agree there shouldn't be a weird user instruction in output, thus have made it empty if not present

EmmaLRussell (approved):
Looks good!

PR #80, opened 2024-10-02
Bacpop-188 Allow species selection for project

absternator:
This is the first part of multi-db work. In subsequent PRs supporting multi db to work
..This PR allows  users to create a project and ensures they select a species.  Follwing is done:
- modal to create project 
- store species when saving project 
- add species to column 
- add species name to project page

Testing: 
- Click on create button on all projects view... then ensure can create new project. see species name and go back home and see this column filled in correctly. 

image
image
image


app/client-v2/src/components/HomeView/CreateProjectButton.vue
```diff
@@ -0,0 +1,92 @@
...
+import { useRouter } from "vue-router";
+
+const props = defineProps<{
+  projects: ProjectOverview[];
+}>();
+
+const { showErrorToast } = useToastService();
+const router = useRouter();
+const apiUrl = getApiUrl();
+const visible = ref(false);
```
EmmaLRussell, line 16:
```suggestion
const dialogVisible = ref(false);
```
Thought this was for the button visibility at first!

app/client-v2/src/components/HomeView/CreateProjectButton.vue
```diff
@@ -0,0 +1,92 @@
...
+  species: "",
+  creation: ""
+});
+
+const isDuplicateName = (name: string) => props.projects.some((project) => project.name === name);
+
+const closeDialog = () => {
+  visible.value = false;
+  species.value = null;
+  errors.name = errors.species = errors.creation = name.value = "";
```
EmmaLRussell, line 30: This is a little bit gruesome. Could use this pattern of having an initial error state which is used for both initialisation and reset on dialog close: https://stackoverflow.com/questions/61184749/how-to-properly-reset-vue-composition-apis-reactive-values

EmmaLRussell (commented):
Looks good to me. I think it would be nice if it remembered your last species choice and populated that in the dialog so you didn't have to choose it every time, but that's a nice to have (for another ticket if at all). 

I don't really think you need the info box with the species name on the project page when the instruction already mentions the species name ("Upload genome sequences for..") Seems a bit redundant to have both,

PR #81, opened 2024-10-03
Bacpop-194 All Beebop changes for multi-db

absternator:
This PR covers all changes that will be needed for multi-db.When the beebop_py portion of work is done, this  can be then merged and tested with that branch. Currently I ve tested with  [beebop_py poc](https://github.com/bacpop/beebop_py/pull/43). Feel free to look for context and play around (API branch is pointing to that poc branch).

 The following is done in this PR 
- Get species info with kmers from API 
- pass species info when running poppunk 
- remove DB_LOCATION related config


app/server/src/controllers/indexController.ts
```diff
@@ -109,5 +107,14 @@ export default (config) => {
                     handleAPIError(request, response, error);
                 });
         },
+
+        async getSketchKmerArguments(request, response) {
```
EmmaLRussell, line 111: Might be less confusing to call this `getSpeciesConfig` - that seems to be the terminology elsewhere in the chain, and other config might get added in future.

EmmaLRussell (commented):
Code looks good to me. 

AMR for strep samples always shows as "unknown" - is that expected? This is on running this branch with the `run_test` script. 

One thing I forgot to ask in the last PR - will the code be backward compatible with old projects stored without a "species" field. Looks like it probably default to undefined, but should default to pneumo: https://github.com/bacpop/beebop/pull/80/files#diff-6b11b12f8cdd9323969a2b053f174be528aebc9603ca7ef6c174af538c441534R104

app/server/src/controllers/indexController.ts
absternator: good catch i missed this in renaming haha

absternator:
> AMR for strep samples always shows as "unknown" - is that expected? This is on running this branch with the `run_test` script.
> 
> One thing I forgot to ask in the last PR - will the code be backward compatible with old projects stored without a "species" field. Looks like it probably default to undefined, but should default to pneumo: https://github.com/bacpop/beebop/pull/80/files#diff-6b11b12f8cdd9323969a2b053f174be528aebc9603ca7ef6c174af538c441534R104

1. Yip I think it is expected, but nick said to just run it anyways initially. But Il follow up on it with him
2. oh yes good catch, I will make the default to pneumo as suggested

EmmaLRussell (approved):
Nice - species name in bold looks good!

PR #82, opened 2024-10-10
Mrc-198 Bug fix microreact dialog bug

absternator:
This PR fixes bug where in a project if the first sample failed and other samples passed after run, then the settings cog on the microreact cog would not open. This was because the dialog was only rendered if the first sample had a cluster assigned. This has now been updated so if any sample has a cluster assigned then the cog opens up dialog.

This is the cog
image


EmmaLRussell (approved):
Works for me! Though tbh I'm a bit confused about why we need to send a cluster id in order to post a token. That's a requirement of the Microreact API?

PR #83, opened 2024-11-08
Bacpop 201 stream microreact

absternator:
This PR allows streaming of microreact jobs per cluster as they come in... The main changes are as below:

- add new microreactClusters status which is a map of clusters and their status.. This is used to track whether a clusters microreact is ready or not.

Testing:

Best to test by running app with samples and seeing that microreact become available as they are finished...

### DIFF
--- a/app/client-v2/src/utils/projectStatus.ts
+++ b/app/client-v2/src/utils/projectStatus.ts
@@ -20,8 +20,9 @@ export const hasMicroreactClusterFailed = (
   status: AnalysisStatus | undefined,
   cluster: string | undefined
 ): boolean => {
-  if (!cluster && Object.values(status?.microreactClusters || {}).some((status) => status == "finished")) return true;
-
+  // occurs when another sample has passed but this one fails
+  if (!cluster && isAnyMicroreactFinished(status?.microreactClusters)) return true;
+  
   if (status?.microreact == "failed" || (cluster && status?.microreactClusters?.[cluster] === "failed")) return true;
 
   return false;
@@ -36,4 +37,10 @@ export const getMicroreactClusterStatus = (
   }
 
   return status?.microreact || "waiting";
+};
+
+export const isAnyMicroreactFinished = (
+  microreactClusterStatuses: Record<string, StatusTypes> | undefined
+): boolean => {
+  return Object.values(microreactClusterStatuses || {}).some((status) => status == "finished");
 };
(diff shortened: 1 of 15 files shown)


app/client-v2/src/utils/projectStatus.ts
```diff
@@ -1,7 +1,50 @@
...
+    return microreactClusterStatuses[cluster] === "finished";
+  }
+  return false;
+};
+
+export const hasMicroreactClusterFailed = (
+  status: AnalysisStatus | undefined,
+  cluster: string | undefined
+): boolean => {
+  // occurs when another sample has passed but this one fails
```
EmmaLRussell, line 23: Do you mean when it's failed to assign a cluster, hence it being falsy?
```suggestion
  // occurs when this sample could not be assigned to a cluster but at least one other sample was assigned
```

Should we show the microreact job as "failed" in that case? It's more like we couldn't run it because a prerequisite failed. How about "cancelled"?

app/client-v2/src/utils/projectStatus.ts
```diff
@@ -1,7 +1,50 @@
...
+  cluster: string | undefined
+): StatusTypes => {
+  if (cluster && status?.microreactClusters?.[cluster]) {
+    return status.microreactClusters[cluster];
+  }
+
+  return status?.microreact || "waiting";
+};
+
+export const isAnyMicroreactQueued = (microreactClusterStatuses: Record<string, StatusTypes> | undefined): boolean => {
```
EmmaLRussell, line 42: Slightly misleading method name, since these could all be completed jobs, not currently queued.
```suggestion
export const haveAnyMicroreactBeenQueued = (microreactClusterStatuses: Record<string, StatusTypes> | undefined): boolean => {
```
Something like that?

EmmaLRussell (commented):
Looks great!

I can run this with the newer sample strep samples you sent over a while ago, but failing with the old fasta examples. Is that expected?

EmmaLRussell:
OK, this now seems to be running fine for me with the original fasta files. Maybe I just made a project with the wrong species before!

app/client-v2/src/utils/projectStatus.ts
absternator: yup cancelled does make sense here.. ive made ticket as this will need a bit more changes https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-204/When-assign-fails-change-network-microreact-to-cancelled-not-failed

PR #84, opened 2024-11-21
Bacpop 186 v9 db

absternator:
The only change required is to update the keys the cytoscape graph points to, as these have changed due to using another library to manage the graphmls.

Testing:
This branch and associated beebop_py branch is deployed to [beebop-dev](https://beebop-dev.dide.ic.ac.uk)

Note: tests failing due to network failing on refs db. working with nick to get fixed.

Note: need to wait for beebop_py to merge first before this merge.

image


absternator:
DO NOT MERGE until beebop_py merged... ALSO revert branch to main after

app/server/jest.config.js
```diff
@@ -2,6 +2,6 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-  testTimeout: 120000,
+  testTimeout: 240000,
```
EmmaLRussell, line 5: oh dear :laughing:

EmmaLRussell (approved):
Looks good. As discussed, it's odd to have to change the graph keys, but since we're the only ones using the subgraphs it shouldn't affect the full graphml files that people might already be using.

app/server/jest.config.js
absternator: that was a test haha good catch I've reverted

PR #86, opened 2024-12-06
Cache Beebop databases in CI

absternator:
This PR does work to allow caching of the beebop databases. The playwright CI and jest integration CI both require pulling databases which can take a  long time. These don't change thus we are now caching, so every subsequent run we don't have to redownload databases.

To allow caching having added `mount` cli option to `run_tests` that allows to mount GHA directory into the `/storage` folder (where dbs are downloaded)

Testing: can re-run pipeline jest CI / unit_and_integration_backend (20.x) (pull_request) and you will see in logs of _run all components _ step that **Database already exists** for the multiple databases


README.md
```diff
@@ -77,6 +77,8 @@ You can also run everything outside pm2, by separately running:
 - `./scripts/run_server`
 - `./scripts/run_client`
 
+*Note: If you wish to override the volune with a custom bind mount pass in -mount {MOUNT_NAME} into `run_test` script*
```
EmmaLRussell, line 80:
```suggestion
*Note: If you wish to override the storage volume with a custom bind mount pass in -mount {MOUNT_NAME} into `run_test` script*
```

scripts/run_dependencies
```diff
@@ -4,32 +4,43 @@ set -ex
...
+  case "$1" in
+    -mount)
+      MOUNT="$2"
+      shift 2
+      ;;
+    *)
+      shift
+      ;;
+  esac
+done
```
EmmaLRussell, line 18: this is shared logic with run _test to find the mount arg - could pull this out into a shared script - or just use a positional arg if it's easier! But it's quite nice and clear to have a named arg.

EmmaLRussell (commented):
Nice!

So the database ends up in the cache because the first time it runs, the script will write it out to the host location of the bind mount?

EmmaLRussell (approved):

PR #87, opened 2025-01-16
Bacpop-209 Beebop:AMR info to Microreact

absternator:
This PR is in conjunction with [beebop_py pr](https://github.com/bacpop/beebop_py/pull/57).
The purpose is to pass along AMR information to an info_csv metadata file that is passed into visualise for microreact. This means that this AMR info will show on the table when visiting microreact page.

Also added Status column as default into microreact and sorted so query samples show first before ref samples. 

Testing:
Create a project with samples and run. Then visit microreact, the AMR columns should show by default and query samples should be on top of ref samples. 
have also deployed on [beebop-dev](https://beebop-dev.dide.ic.ac.uk/) if its easier

### DIFF
--- a/app/client-v2/src/types/projectTypes.ts
+++ b/app/client-v2/src/types/projectTypes.ts
@@ -23,6 +23,17 @@ export interface AMR {
   species: boolean;
 }
 
+export interface AMRForCsv {
+  "Penicillin Resistance": string;
+  "Chloramphenicol Resistance": string;
+  "Erythromycin Resistance": string;
+  "Tetracycline Resistance": string;
+  "Cotrim Resistance": string;
+}
+
+export interface AMRMetadataCsv extends AMRForCsv {
+  ID: string;
+}
 export interface ProjectSample {
   hash: string;
   filename: string;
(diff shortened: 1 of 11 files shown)


app/client-v2/src/types/projectTypes.ts
```diff
@@ -23,6 +23,17 @@ export interface AMR {
...
+export interface AMRForCsv {
+  "Penicillin Resistance": string;
+  "Chloramphenicol Resistance": string;
+  "Erythromycin Resistance": string;
+  "Tetracycline Resistance": string;
+  "Cotrim Resistance": string;
+}
+
+export interface AMRMetadataCsv extends AMRForCsv {
+  ID: string;
```
EmmaLRussell, line 35: ..deviating from more customary lower case `id`..?

EmmaLRussell (approved):
Should we flag up to Nick/users that the column names in the export file have changed?

absternator:
> Should we flag up to Nick/users that the column names in the export file have changed?

good idea will do

PR #88, opened 2025-02-04
Bacpop 208 nw mc 1 call

absternator:
The following PR is done along with the beebop_py code here and should be tested together https://github.com/bacpop/beebop_py/pull/59

This PR accommodates for the fact that now there is only 2 steps... assign and visualise(network + microreact combined).
Note: Network tab will only become available once all networks are loaded, this is for performance reasons as we display them all on 1 page. Have made ticket to revisit this however --> https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-212/Revist-network-display

Testing:
Test with frontend PR. Test networks and microreact work fine, and the zip downloads do as well.


EmmaLRussell (approved):
This looks good to me. 

Arguably, it's a bit misleading now to have separate statuses in the Network and Microreact columns, but I don't think that's a big issue. We want those columns to be separate anyway as they have different downloads.

PR #89, opened 2025-03-14
Network library change updates

absternator:
This is the corresponding changes to [beebop_py pr](http://localhost:5173/project/ec3fa538308bf6c7fc0e617b28d01bce). This is needed as new graph library graph-tool produces slightly different key names.

Testing: 
test with the beebop_py branch by setting API_BRANCH=speedup-subgraph-creation in common.
The network graph should still be working and not broken

### DIFF
--- a/app/client-v2/src/components/ProjectView/CytoscapeCanvas.vue
+++ b/app/client-v2/src/components/ProjectView/CytoscapeCanvas.vue
@@ -29,14 +29,14 @@ onMounted(async () => {
         style: {
           width: "10px",
           height: "10px",
-          content: "data(d1)",
+          content: "data(key2)",
           "font-size": "7px",
           color: "#00CC66",
           "background-color": "rgba(45, 212, 191, 0.44)"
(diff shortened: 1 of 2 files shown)


EmmaLRussell (commented):
Works for me!

I always get caught out when I try to run beebop from sources after I haven't for a while, and try to run `./scripts/run_docker` - which doesn't work because we're not pushing SHA images any more. We should probably either remove that script or update it to use the branch images. Want to make a ticket for that?

absternator:
> I always get caught out when I try to run beebop from sources after I haven't for a while, and try to run `./scripts/run_docker` - which doesn't work because we're not pushing SHA images any more. We should probably either remove that script or update it to use the branch images. Want to make a ticket for that?

good idea have made ticket https://mrc-ide.myjetbrains.com/youtrack/issue/bacpop-215/Remove-restructure-rundocker-script

EmmaLRussell (approved):

PR #94, opened 2025-09-15
Dont show failed on cluster when unable to assign to cluster and novel genotype

absternator:
The following PR contains update when cluster assignment fails due to not bein able to be asssigned to database (most likely novel genotype), then show warning tag rather than failed. This is requested by nick and i have shown him proposed UI.

Testing:
i can quickly show on my computer... or i have sent sample to you and you can run project Streptococcus pneumoniae and see result as per screenshot

image

image

### DIFF
--- a/app/client-v2/src/__tests__/components/ProjectView/ProjectPostRun.spec.ts
+++ b/app/client-v2/src/__tests__/components/ProjectView/ProjectPostRun.spec.ts
@@ -248,7 +248,7 @@ describe("RunProject", () => {
     expect(screen.queryByText(/network graphs/i)).not.toBeInTheDocument();
   });
 
-  it.only("should show contact tag when sample failType is warning", async () => {
+  it("should show contact tag when sample failType is warning", async () => {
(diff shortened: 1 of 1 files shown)


EmmaLRussell (commented):
I see what you mean about the hackiness of the implementation which has been forced onto you by the response from beebop_py. It does feel a bit shaky to rely on that fail reason description string. 

I think if you were going to do it property, the "failReason" coming back from beebop_py would be an object rather than just a string, and it could have a "failType" which could be an enum  type with UNEXPECTED_FAILURE or POTENTIAL_NOVEL_GENOTYPE and then you could switch on that when deciding how to display it. 

I also don't think the tooltip is an ideal way to present an email address - it's currently wrapping and isn't clickable. Again, if you had failure type in the response, you could potentially keep the "Contact.." user string and email address in the front end and use the type to switch whether to display that e.g. could have a "More info and help" link under the "not in database" slug, which could open a little modal with the mailto link. Or could you have a more structured popover rather than text only tooltip?

PR #99, opened 2026-02-13
Fix: Sublineage not showing on csv download

absternator:
There was issue that sublinages was not showing on csv download. this was because we use the first sample for all headers. so if there is missing sublineage for that field then then the csv would be missng header thus field... Thus now checking all samples for headers. 

Also have it to show empty string if field not exists. this is much cleaner than old 'undefined' that used to show

Testing:
add samples and run. then run export as see sublinages and if any missing then will show empty string... Try `S_pneumoniae_Hungary19A_6_GCF_000019265_1.fa and S_pneumoniae_M26368_GCF_003351725_1.fa` (no sublinage) + `S_pneumoniae_D39V_GCF_003003495_1.fa` ... message me if you dont have these samples

### DIFF
--- a/app/client-v2/src/__tests__/utils/projectCsvUtils.spec.ts
+++ b/app/client-v2/src/__tests__/utils/projectCsvUtils.spec.ts
@@ -87,6 +87,32 @@ describe("projectCsvUtils", () => {
 
       expect(result).toBe("");
     });
+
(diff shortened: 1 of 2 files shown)


EmmaLRussell (approved):
LGTM. I suppose you could consider putting "none" or "unavailable" (which is shown in the app) rather than empty strings - but I think those are fine too.

